PR #82, opened 2021-04-02
Introducing the concept of feature events to export data

### LINKED ISSUE #70: (feature) Add a way to collect decision in variation
# Requirements
We want a way to collect the data of which user is in which variation.

This could be written in different formats (webhook, json, csv, s3, locally ...).
Since it could represent a lot of data, we want to avoid writing data every time we have a call to the variation because it will put a lot of load on the webhook to collect the affectation choices.
We should have a mechanism that collects all the variation for the last X seconds and call the webhook with all the affectation in that period in bulk.

We should also put a limit on the number of events we get before sending the bulk.
Something like _call the webhook every X seconds or if the number of events equals X_.

The number of seconds and number of seconds should be a parameter.
The webhook configuration should be consistent with the webhook configuration for change in flags ([see doc](https://github.com/thomaspoignant/go-feature-flag#webhook)).

thomaspoignant:
# Description
Introducing the concept of data collector.
This allows to store and send to a third party all the results of the `variations` call.

It means that you can see exactly what was the value used for a flag during the evaluation.
With these data, you have the exact context of your evaluation and it can help you to have the exact context of what happened.
You can also use these data to join them with other metrics to see if one version of the flag performs better than another one.

# Changes include

# Closes issue(s)
Resolve #70 

# Checklist
- [x] I have tested this code
- [x] I have updated the Readme


thomaspoignant:
This PR is almost ready, this version contains only exports in **local file**, but it will be easy to extend this by adding more locations to export the data (see #87 or #88).

The documentation is missing and, should be updated.

PR #246, opened 2022-05-31
Add environment parameter in config

### LINKED ISSUE #175: (feature) Add environment parameters in the config
# Requirements
Add an environment parameter in the config.
This environment field should be added to all users in the param so we can add rules based on the environment and have the same config file for all your environment.

mtwstudios:
# Description
Checking the current environment / flagging by environment is a common use cases for projects I'm involved in, so this adds environment awareness to feature flags. You can set environment as a field in the Config passed to Init, then check them in rule (e.g. `rule: env == "dev"`)

I realize I need to address tests/docs, but wanted to get feedback on the approach first to see if it was acceptable.

# Changes include

# Closes issue(s)

Resolve #175

# Checklist
- [x] I have tested this code
- [x] I have updated the documentation (README.md and /docs)


internal/flagv1/flag_data.go
```diff
@@ -77,9 +78,8 @@ func (f *FlagData) Value(flagName string, user ffuser.User) (interface{}, string
 
 func (f *FlagData) isExperimentationOver() bool {
 	now := time.Now()
-	return f.Rollout != nil && f.Rollout.Experimentation != nil && (
-		(f.Rollout.Experimentation.Start != nil && now.Before(*f.Rollout.Experimentation.Start)) ||
-			(f.Rollout.Experimentation.End != nil && now.After(*f.Rollout.Experimentation.End)))
+	return f.Rollout != nil && f.Rollout.Experimentation != nil && ((f.Rollout.Experimentation.Start != nil && now.Before(*f.Rollout.Experimentation.Start)) ||
```
thomaspoignant, line 81: can we keep the same indentation as before to avoid having a too-long line?

internal/flagv1/flag_data.go
```diff
@@ -2,13 +2,14 @@ package flagv1
...
-	"github.com/thomaspoignant/go-feature-flag/ffuser"
-	"github.com/thomaspoignant/go-feature-flag/internal/utils"
 	"math"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/nikunjy/rules/parser"
+	"github.com/thomaspoignant/go-feature-flag/ffuser"
+	"github.com/thomaspoignant/go-feature-flag/internal/utils"
```
thomaspoignant, line 12: Can we keep the import as before, there is no change in them just a different order.

internal/flagv1/flag_data.go
```diff
@@ -113,7 +113,9 @@ func (f *FlagData) evaluateRule(user ffuser.User) bool {
 	}
 
 	// Evaluate the rule on the user.
-	return parser.Evaluate(f.getRule(), utils.UserToMap(user))
+	userMap := utils.UserToMap(user)
+	userMap["env"] = environment
```
thomaspoignant, line 117: Since by default `env` is **empty** I will avoid having the key `env` in the map if there is no value.
```suggestion
	if environment != "" {
		userMap["env"] = environment
	}	
```

What do you think about that?

config.go
```diff
@@ -3,10 +3,11 @@ package ffclient
 import (
 	"context"
 	"errors"
-	"github.com/thomaspoignant/go-feature-flag/ffnotifier"
 	"log"
 	"time"
 
+	"github.com/thomaspoignant/go-feature-flag/ffnotifier"
```
thomaspoignant, line 9: Can we keep the import as before, there is no change in them just a different order.

internal/flagv1/flag_priv_test.go
```diff
@@ -1,10 +1,11 @@
 package flagv1
 
 import (
-	"github.com/stretchr/testify/assert"
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
```
thomaspoignant, line 7: Can we keep the import as before, there is no change in them just a different order.

internal/flagv1/flag_pub_test.go
```diff
@@ -2,11 +2,12 @@ package flagv1_test
 
 import (
 	"fmt"
+	"testing"
```
thomaspoignant, line 5: Can we keep the import as before, there is no change in them just a different order.

variation.go
```diff
@@ -2,6 +2,7 @@ package ffclient
 
 import (
 	"fmt"
+
```
thomaspoignant, line 5: Can we remove this blank line since it is not related to your change

thomaspoignant (changes requested):
Hey, @mtwstudios thanks for your contribution this is awesome.
The approach you have is the one I had in mind when writing the issue so it looks good to me.

Despite that I made some minor changes request in your pull request to avoid changing unused things.

I would also suggest adding some tests where you are testing the environment feature because in your test you are always sending an empty `environment` so it will not test that this is working as expected.

internal/flagv1/flag_data.go
mtwstudios: makes sense to me!

mtwstudios:
Thanks for the feedback @thomaspoignant ! I think I've addressed everything, including the cleanup, and adding tests and docs for the change. Let me know if there's any other changes you think are needed.

mtwstudios:
Hey @thomaspoignant, yep - with this solution, env would be specified in the rule.

The advantage of having it be specified in the rule is that it allows for more flexibility for various use cases. In particular:

- **Enabling for multiple (but not all) environments.** This is a very common use case on teams I've been on - for example to enable a flag in local development environments and staging environments but not in production
  - e.g. `rule: (env == "dev") or (env == "staging")`
 - **Enabling the flag based on certain rules only in certain environments.** For example it's enabled for all users in non-production environments, but in production it's only enabled for specific users
   - e.g. `rule: (env != "prod") or (user_id == 1234)`

I think having `environment` separate from `rule`, while might be easier in the simplest case, prevents it from being used in cases like the examples above. This would make it much less useful/powerful in my opinion.

> My only concern is that we cannot have 2 different configurations for the same flag, and it could be a limitation.

I'm not sure what you mean by this - can you give an example?

thomaspoignant (approved):

PR #253, opened 2022-06-10
Add new GetFlagsFromCache public method

### LINKED ISSUE #251: Make `getFlagFromCache` method public
# Motivation

I'm running `go-feature-flags` in my API and in my `health` endpoint I'd like to return all the `flags` that I have configured and the current version in cache. I see that the `getFlagFromCache` method could give me all the information that I need but it's currently private. Is it ok if we made it public? Happy to send a PR to address that.

marcosnils:
This allows retrieving flag cache metadata like version, rule etc
so the application can use it for visibility purposes. i.e: you want to
know what version of a specific flag your `ffclient` is currently using

resolves #251

### DIFF
--- a/feature_flag_test.go
+++ b/feature_flag_test.go
@@ -1,15 +1,16 @@
 package ffclient_test
 
 import (
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/stretchr/testify/assert"
-	ffclient "github.com/thomaspoignant/go-feature-flag"
-	"github.com/thomaspoignant/go-feature-flag/testutils/mock"
 	"io/ioutil"
 	"log"
 	"os"
 	"testing"
 	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/stretchr/testify/assert"
+	ffclient "github.com/thomaspoignant/go-feature-flag"
+	"github.com/thomaspoignant/go-feature-flag/testutils/mock"
 
 	"github.com/thomaspoignant/go-feature-flag/ffuser"
 )
@@ -66,6 +67,20 @@ func TestValidUseCase(t *testing.T) {
 
 	allFlags := ffclient.AllFlagsState(user)
 	assert.Equal(t, 2, len(allFlags.GetFlags()))
+}
+
+func TestAllFlagsFromCache(t *testing.T) {
+	err := ffclient.Init(ffclient.Config{
+		Retriever:       &ffclient.FileRetriever{Path: "testdata/flag-config.yaml"},
+		PollingInterval: 5 * time.Second,
+	})
+	defer ffclient.Close()
+
+	assert.NoError(t, err)
+	flags, err := ffclient.GetFlagsFromCache()
+
+	assert.NoError(t, err)
+	assert.Len(t, flags, 2)
 }
 
 func TestValidUseCaseToml(t *testing.T) {
@@ -157,7 +172,7 @@ func TestUpdateFlag(t *testing.T) {
   default: false`
 
 	flagFile, _ := ioutil.TempFile("", "")
-	_ = ioutil.WriteFile(flagFile.Name(), []byte(initialFileContent), 0600)
+	_ = ioutil.WriteFile(flagFile.Name(), []byte(initialFileContent), 0o600)
 
 	gffClient1, _ := ffclient.New(ffclient.Config{
 		PollingInterval: 1 * time.Second,
@@ -176,7 +191,7 @@ func TestUpdateFlag(t *testing.T) {
   false: false
   default: false`
 
-	_ = ioutil.WriteFile(flagFile.Name(), []byte(updatedFileContent), 0600)
+	_ = ioutil.WriteFile(flagFile.Name(), []byte(updatedFileContent), 0o600)
 
 	flagValue, _ = gffClient1.BoolVariation("test-flag", ffuser.NewUser("random-key"), false)
 	assert.True(t, flagValue)
@@ -196,7 +211,7 @@ func TestImpossibleToLoadfile(t *testing.T) {
   default: false`
 
 	flagFile, _ := ioutil.TempFile("", "impossible")
-	_ = ioutil.WriteFile(flagFile.Name(), []byte(initialFileContent), 0600)
+	_ = ioutil.WriteFile(flagFile.Name(), []byte(initialFileContent), 0o600)
 
 	gffClient1, _ := ffclient.New(ffclient.Config{
 		PollingInterval: 1 * time.Second,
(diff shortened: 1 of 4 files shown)


marcosnils:
@thomaspoignant WDYT now?

variation.go
```diff
@@ -155,6 +161,13 @@ func (g *GoFeatureFlag) AllFlagsState(user ffuser.User) flagstate.AllFlags {
 	return allFlags
 }
 
+// GetFlagsFromCache returns all the flags present in the cache with their
+// current state when calling this method. If cache hasn't been initialized, an
+// error reporting this is returned.
+func (g *GoFeatureFlag) GetFlagsFromCache() (map[string]flag.Flag, error) {
```
thomaspoignant, line 167: Could you also expose `GetFlagsFromCache()` when using the static version of `go-feature-flag`.

As you can see with other functions, we expose them statically and within the `GoFeatureFlag` struct, you can look at `JSONVariation` for the example.

thomaspoignant (changes requested):
This looks good to me, except that we are missing the static function when using the `Init` function to initialize `go-feature-flag`.

I am also asking myself about the naming `GetFlagsFromCache()` is not super explicit for someone that has never seen the internal of the SDK.

The only name that comes to my mind is `GetFlagsConfiguration()` but I am not 100% sure if it is more understandable. WDYT?

marcosnils:
Hey Thomas! thx  for taking the time to review.

> This looks good to me, except that we are missing the static function when using the `Init` function to initialize `go-feature-flag`.

Not sure what you mean we are missing here.



> I am also asking myself about the naming `GetFlagsFromCache()` is not super explicit for someone that has never seen the internal of the SDK.
> 
> The only name that comes to my mind is `GetFlagsConfiguration()` but I am not 100% sure if it is more understandable. WDYT?

I personally don't see any problems in exposing internals of the library in the function name. In fact, in this case I believe it's better because it's important for the user to understand that this flag configs are coming from the cache and that you could get outdated information if the poll interval still hasn't triggered.

thomaspoignant:
> Not sure what you mean we are missing here.

You have 2 ways to use the SDK, you can have a unique and [static initialization](https://github.com/thomaspoignant/go-feature-flag/blob/main/README.md#getting-started) or you can decide to have[ multiple instance of the SDK](https://github.com/thomaspoignant/go-feature-flag/blob/main/docs/configuration.md#multiple-configuration-flag-files).

It also means that we have 2 different levels for the same function.

If you look at `JSONVariation` for example we have 2 different signatures
- https://github.com/thomaspoignant/go-feature-flag/blob/1a486bfacd0bfd5ba1f279e765cb28d7a722e733/variation.go#L48
- https://github.com/thomaspoignant/go-feature-flag/blob/1a486bfacd0bfd5ba1f279e765cb28d7a722e733/variation.go#L110

We need something similar for the new function.



> I personally don't see any problems in exposing internals of the library in the function name. In fact, in this case I believe it's better because it's important for the user to understand that this flag configs are coming from the cache and that you could get outdated information if the poll interval still hasn't triggered.

You convince me about the naming we can keep `GetFlagsFromCache()`.

marcosnils:
> You have 2 ways to use the SDK, you can have a unique and [static initialization](https://github.com/thomaspoignant/go-feature-flag/blob/main/README.md#getting-started) or you can decide to have[ multiple instance of the SDK](https://github.com/thomaspoignant/go-feature-flag/blob/main/docs/configuration.md#multiple-configuration-flag-files).

gotcha. I'll send the fix in a minute :pray:

feature_flag_test.go
```diff
@@ -176,7 +191,7 @@ func TestUpdateFlag(t *testing.T) {
   false: false
   default: false`
 
-	_ = ioutil.WriteFile(flagFile.Name(), []byte(updatedFileContent), 0600)
+	_ = ioutil.WriteFile(flagFile.Name(), []byte(updatedFileContent), 0o600)
```
thomaspoignant, line 194:
```suggestion
	_ = ioutil.WriteFile(flagFile.Name(), []byte(updatedFileContent), 0600)
```

feature_flag_test.go
```diff
@@ -196,7 +211,7 @@ func TestImpossibleToLoadfile(t *testing.T) {
   default: false`
 
 	flagFile, _ := ioutil.TempFile("", "impossible")
-	_ = ioutil.WriteFile(flagFile.Name(), []byte(initialFileContent), 0600)
+	_ = ioutil.WriteFile(flagFile.Name(), []byte(initialFileContent), 0o600)
```
thomaspoignant, line 214:
```suggestion
	_ = ioutil.WriteFile(flagFile.Name(), []byte(initialFileContent), 0600)
```

feature_flag_test.go
```diff
@@ -264,7 +279,7 @@ func TestFlagFileUnreachable(t *testing.T) {
 	flagValue, _ := gff.StringVariation("test-flag", ffuser.NewUser("random-key"), "SDKdefault")
 	assert.Equal(t, "SDKdefault", flagValue, "should use the SDK default value")
 
-	_ = ioutil.WriteFile(flagFilePath, []byte(initialFileContent), 0600)
+	_ = ioutil.WriteFile(flagFilePath, []byte(initialFileContent), 0o600)
```
thomaspoignant, line 282:
```suggestion
	_ = ioutil.WriteFile(flagFilePath, []byte(initialFileContent), 0600)
```

thomaspoignant (commented):
I see a lot of changes in the code style, order of import, etc ...
I am not sure this is really needed and it brings noises to your PR.

Do you mind changing only what is needed?
If you prefer I can help with that.

marcosnils:
> I see a lot of changes in the code style, order of import, etc ...
> I am not sure this is really needed and it brings noises to your PR.

my vim LSP applied them automatically since I'm using https://github.com/mvdan/gofumpt. Do you still prefer me to revert them? gofumpt is being widely adopted in the Go ecosystem.

thomaspoignant:
> > I see a lot of changes in the code style, order of import, etc ...
> > I am not sure this is really needed and it brings noises to your PR.
> 
> my vim LSP applied them automatically since I'm using https://github.com/mvdan/gofumpt. Do you still prefer me to revert them? gofumpt is being widely adopted in the Go ecosystem.

no we will keep the change, I will just add `gofumpt` into the configuration of `golangci-lint` for future releases

thomaspoignant:
@marcosnils thanks a lot for your contribution.
I will release a new version Today with this new feature.

marcosnils:
sure Thomas! glad to help a tiny bit to make gff great! Please to collaborate with you and thx for the pointers. 

Keep up the awesome work! I know by experience that doing OSS is not an easy task.

thomaspoignant:
Thank @marcosnils.
I am already in contact with the people at @open-feature, you can find me in the [interested parties](https://github.com/open-feature/community/blob/main/interested-parties.md) list 😉

thomaspoignant:
Version `v0.24.0` contains this change.

PR #264, opened 2022-07-04
Move retriever / notifier / exporter on their own packages.

### LINKED ISSUE #256: (change) Move retrievers into sub-pacakges (or external libs?)
# Motivation

The go-feature-flag library provides a nice and hopefully growing list of retrievers. Unfortunately all retrievers are made available in the top-level package, which means that all projects that do depend on go-feature-flag automatically pull in and compile all those dependencies, no matter what. With the recent addition of the kubernetes client the compile, lint times, and binary file sizes have increased quite significantly. In my case each each have doubled the least... and my local pre-commit hook running golangci-lint even went up to 200s.

# Requirements

Users of go-feature-flag should be allowed to pull in only the dependencies they actually need. This can be achieved by moving the retrievers into their own packages like `go-feature-flag/retrievers/...` for example, or moving them into separate Repositories (which might be annoying maintenance wise...).

thomaspoignant:
# Description

## :warning: Breaking changes  :warning:

This version change the way to do your SDK initialization in particular when it comes to `notifier`, `provider` and `exporter`.

### Why are we doing this breaking changes?

This change is needed because we are adding more and more retriever / exporter and notifier, and with that we are also adding more dependencies to the project.  
And since all the dependencies are in the same package we were increasing the size of your build even if you were not using the new providers.

The best example is when we introduce the kubernetes provider, it has added **~20Mo** in your application because the `client-go` from kubernetes is huge. 

## What did we change?
We have moved all retriever / exporter and notifier in a dedicated package for each of them.  
With this new organisation when building your app we will use only the dependencies related to your configuration.

We also change the way to deal with the `notifiers` to align it with the way it work for other extensions.

### How to migrate?

Edit your init function of `go-feature-flag` and replace your retriever, exporter, notifier following this.

#### Retrievers
```go
ffclient.FileRetriever          ->      fileretriever.Retriever
ffclient.GithubRetriever        ->      gcstorageretriever.Retriever
ffclient.GithubRetriever        ->      githubretriever.Retriever
ffclient.HTTPRetriever          ->      httpretriever.Retriever
ffclient.KubernetesRetriever    ->      k8sretriever.Retriever
fflcient.S3Retriever            ->      s3retriever.Retriever
```

#### Exporters
```go
ffexporter.File                 ->      fileexporter.Exporter
ffexporter.GoogleCloudStorage   ->      gcstorageexporter.Exporter
ffexporter.Log                  ->      logsexporter.Exporter
ffexporter.S3                   ->      s3exporter.Exporter
ffexporter.Webhook              ->      webhookexporter.Exporter
```

#### Notifiers

Since in this PR we change the way to work with notifier, you have more impact when configuring them.

Before this PR to configure a notifier you had something like that.

```go
_, err := ffclient.New(ffclient.Config{
    // ...
    Notifiers: []ffclient.NotifierConfig{
        &ffclient.WebhookConfig{
            // ...
        },
    },
})
```

With this PR we had remove the `ffclient.NotifierConfig` struct to use directly the `notifier` them self.  
It means that now `ffclient.Config.Notifiers` has the type `[]notifier.Notifier`.  
So it will look like:

```go
_, err := ffclient.New(ffclient.Config{
    // ...
    Notifiers: []notifier.Notifier{
        &webhooknotifier.Notifier{
            // ...
        },
    },
})
```

To follow the same pattern has the `retrievers` and `exporters` we also have changed the struct you were using.

```go
notifier.LogNotifier        ->      logsnotifier.Notifier
ffclient.SlackNotifier      ->      slacknotifier.Notifier
ffclient.WebhookConfig      ->      webhooknotifier.Notifier
```

# Changes include
- [x] Breaking changes (change that is not backward-compatible and/or changes current functionality)

# Closes issue(s)
Resolve #256

# Checklist
- [x] I have tested this code
- [x] I have updated the documentation (README.md and /docs)


urso:
Thank you, the PR description is very detailed.

I only skimmed over the changes, but LGTM. Only nit is that the placing of the testdata feels a little non-idiomatic. As the exporters/retrievers/notifiers are more isolated now I would normally expect the testdata right next to each exporter. Like `exporter/file/testdata` and so on.

Issue #310, opened 2022-08-20
(bug) Google cloud storage retriever only gets the file in the first attempt


jferrl:
### Observed behavior

It was not possible to use an instance of `gcstorageretriever.Retriever{}` since it is only able to get the file on the initial bootstrap of go-feature-flag instance. After feature flag client is up and running, the internal flags cache is no longer updated when **gcstorageretriever** is used.

### Expected Behavior

`gcstorageretriever.Retriever{}` should be able to the flags when the file stored in google cloud storage changes.

### Steps to reproduce

Using the following code(just an exaple): 

```go
ffcli, err := ffclient.New(ffclient.Config{
  PollingInterval: 3 * time.Second,
  Retriever: &gcstorageretriever.Retriever{
      Options: []option.ClientOption{option.WithoutAuthentication()},
  Bucket: "my-bucket",
  Object: "flags.yaml",
  }
})
defer ffcli.Close()
```

After the bootstrap, we are able to get current flags data, for example using the following func:

```go
flags := ffcli.AllFlagsState(user)
```

But, if the file is updated in google cloud storage, go-feature-flag is not able to retrieve the current config.

Moreover, We can see several log traces that refer to the following line of code(saying file not found)
```go
fmt.Errorf("unable to read from GCP Object %s in Bucket %s, error: %s", retriever.Bucket, retriever.Object, err)
```

thomaspoignant:
@jferrl thanks for reporting this issue.

I am sorry that it is not working as expected 😓
I am currently on PTO so I am not able to check this before next week.

In the mean time if you want to work on resolving this issue it will be much appreciated.

jferrl:
> @jferrl thanks for reporting this issue.
> 
> I am sorry that it is not working as expected 😓 I am currently on PTO so I am not able to check this before next week.
> 
> In the mean time if you want to work on resolving this issue it will be much appreciated.

No worries! I'm working on the fix! 
Thanks! 😄

thomaspoignant:
Resolves by #311

PR #311, opened 2022-08-20
Fix/gcs retriever

jferrl:
# Description

## What was the problem?
It was not possible to use an instance of gcstorageretriever.Retriever{} since it is only able to get the file on the initial bootstrap of go-feature-flag instance. After feature flag client is up and running, the internal flags cache is no longer updated when gcstorageretriever is used.

## How it is resolved?

The issue has been resolved avoiding the use of the file reader as a field of `Retriever`. So, after the first bootstrap of `Retriever` the file reader was closed because the code closes the reader after done reading. If we try out another call to Retriever.Retrieve() the reader is not nil and it was closed, so that, it was not able to get the updated file. In that case, we could see some log trace like:
```go
fmt.Errorf("unable to read from GCP Object %s in Bucket %s, error: %s", retriever.Bucket, retriever.Object, err)
```

## How can we test the change?
- Unit tests
- Example script using `gcstorageretriever.Retriever{}` 

# Changes include

# Closes issue(s)
https://github.com/thomaspoignant/go-feature-flag/issues/310

# Checklist
- [ ] I have tested this code
- [ ] I have updated the documentation (README.md and /docs)


thomaspoignant:
Don't worry about the linting issue, I will fix it in another PR and will works for you.

thomaspoignant (approved):

jferrl:
@thomaspoignant PR able to merge! 🚀

thomaspoignant (approved):
Thanks a lot for you contribution @jferrl 
Sorry for the delay in the review.

I will release a new version with this fix today.

PR #425, opened 2022-12-05
Fix race on close and overwriting Once

urso:
# Description

> What was the problem?

Having multiple instances of feature flag (in parallel tests) did lead to race conditions. Turns out that `Close` did overwrite the global `onceFF` variable. As rule of thumb: do never reset `onceFF`. Besides the race on `Close` there is the chance of racing with `Init` overwriting the global `ff` while the current one is closed.

> How it is resolved?

Let's use a mutex in the global `Init` and `Close` functions. These functions should not be used all that often anyways. No need trying to optimize it.

> If there are breaking changes, please describe them in detail and why we cannot avoid them.

If there is an active `ff` we do error out on `Init`. We can easily restore the old behavior and just return the current instance. But the config object is rather 'complex' and I wanted to ensure that callers of `Init` really get the active FF instance they have configured or erroring out if that is not possible.

It is no change we can not avoid, just felt semantically more correct to me.

> How can we test the change?

Create many GoFeatureFlag objects using `New` and try to close them all concurrently. The test would require the race detector to be enabled. Although we did not create a global instance, all did try to reset the global `onceFF`. 

# Changes include
~~- [ ] New feature (non-breaking change that adds functionality)~~
- [x] Breaking changes (change that is not backward-compatible and/or changes current functionality)

# Closes issue(s)

Resolve #

# Checklist
- [ ] I have tested this code
- [ ] I have updated the documentation (README.md and /docs)


thomaspoignant:
Hey @urso, thanks a lot for raising this pull request.
Sorry for the delay, I will review this PR Today.

thomaspoignant:
@urso, reading your issue I think we can do something easier than using a `mutex`.
I would move the part of code doing `onceFF = sync.Once{}` into the `func Close()` directly so we stop having the issue when we try to update multiple time the `onceFF` variable.

Our close functions will look like this now:

```go
// Close the component by stopping the background refresh and clean the cache.
func Close() {
	onceFF = sync.Once{}
	ff.Close()
}

//...

// Close wait until thread are done
func (g *GoFeatureFlag) Close() {
	if g != nil {
		if g.cache != nil {
			// clear the cache
			g.cache.Close()
		}
		if g.bgUpdater.updaterChan != nil && g.bgUpdater.ticker != nil {
			g.bgUpdater.close()
		}

		if g.dataExporter != nil {
			g.dataExporter.Close()
		}
	}
}
```

Do you think this could fix your issue?

PR #512, opened 2023-02-15
Add lint command

### LINKED ISSUE #511: (feature) Configuration linter
### Requirements

Hi,

I'd like to add a new command to lint flag files, which could be used in CI systems to ensure that changes made to configuration files can be parsed by go-feature-flag.

Is this something that's been considered before? I searched for "lint", "linter" but didn't find anything relevant.

Here's how I think it could work:

    go-feature-flag-lint --flags-file <path to flags file>

The exit code would be non 0 if there are errors:
* non-valid json, toml, or yaml
* query formatting errors
* etc…

What do you think?

garriguv:
# Description

This PR adds a new lint command to verify that flag files can be parsed by GO Feature Flag.

The command will exit with a non-zero code if any errors are detected. Individual errors are printed out to stderr.

It's pretty barebones but does the job well. We could consider adding different output formats like json to enable easier processing of the errors, but that would require a little more design consideration.

It's been a while since I wrote any Go, so don't feel like you have to be nice in the comments 😉

Returning `[]error` from `Lint` doesn't feel idiomatic at all. I've considered bringing a dependency like https://github.com/hashicorp/go-multierror but I'd like to get your opinion first.
We could also create a struct to store errors and return that instead. Something like:

```go
type LintError struct {
	File string
	Err  error
}
```

Do you have a preference on where documentation should be written on the website?

# Changes include

- [x] Adds a new `lint` command

# Closes issue(s)

Resolve #511 

# Checklist
- [X] I have tested this code

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -50,6 +50,28 @@ builds:
         goarch: arm
         goarm: 6
 
+  - id: go-feature-flag-lint-cli
+    main: ./cmd/lint
+    binary: go-feature-flag-lint-cli
+    goos:
+      - darwin
+      - linux
+      - windows
+    goarch:
+      - 386
+      - amd64
+      - arm64
+      - arm
+    goarm:
+      - 6
+      - 7
+    ignore:
+      - goos: darwin
+        goarch: 386
+      - goos: linux
+        goarch: arm
+        goarm: 6
+
 archives:
   - id: go-feature-flag-migration-cli
     name_template: go-feature-flag-migration-cli_{{ .Version }}_{{ .Os }}_{{ .Arch }}{{ with .Arm }}v{{ . }}{{ end }}{{ with .Mips }}_{{ . }}{{ end }}{{ if not (eq .Amd64 "v1") }}{{ .Amd64 }}{{ end }}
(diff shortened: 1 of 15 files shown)


thomaspoignant (commented):
I have read your PR and it seems fine to me, it is an excellent addition.
I don't see any issue with adding a new dependency so feel free to do it if you want to.

If we want to make this command line even better we can try to convert the flag and call the function [`IsValid`](https://github.com/thomaspoignant/go-feature-flag/blob/a4c9c5065a8bbbd4616a490ef2328c8d16565b8d/internal/flag/internal_flag.go#L178) for each flag, we will get an even better linter.

garriguv:
@thomaspoignant cool, thank you for the review! Let's not pull the 3rd party dependency at this stage. I don't think it's worth it.

> If we want to make this command line even better we can try to convert the flag and call the function [IsValid](https://github.com/thomaspoignant/go-feature-flag/blob/a4c9c5065a8bbbd4616a490ef2328c8d16565b8d/internal/flag/internal_flag.go#L178) for each flag, we will get an even better linter.

I'm already doing that [here](https://github.com/thomaspoignant/go-feature-flag/pull/512/files#diff-f9af5b1f03ce892bc6f0fe816bcddbb0b2bbc5f49a4edfefdf29250b86d79952R42-R45). Is this what you meant?

What's the next step here? I'm up for adding documentation on the website, should it be top-level after _Migrate from v0.x.x to v1.x.x_?

thomaspoignant:
Oh sorry about this IsValid thing I read it to fast.
Yes the only remaining things will be the doc.

And yes you could put it in the same level as the migration cli.

thomaspoignant (approved):
I think this is good to be merged.
We can work on the documentation in a separated pull request (I can do it if you want).

garriguv:
Sounds good 👍 I'll open a PR with the documentation tonight. It'll feel good to have done the whole thing!

PR #528, opened 2023-02-22
Add GitHub action documentation (lint)

garriguv:
# Description

Thank you for adding the `lint` documentation! I was waiting to add it to our repository to write it, but you had already done it.

I think it might be useful to show people how to use it in GitHub actions, so here it is.


thomaspoignant (approved):
Yes the doc is even better like this thanks a lot 🙏

PR #531, opened 2023-02-22
create relay-proxy helm chart

### LINKED ISSUE #530: (feature) Supports Go Feature Flag relay proxy deployment to Kubernetes
### Requirements

To support Go Feature Flag relay proxy deployment to Kubernetes, it will be nice to provide out-of-box K8s deployment script or Helm chart. Currently relay proxy docker image documentation (https://hub.docker.com/r/thomaspoignant/go-feature-flag-relay-proxy) only covers docker environment.

thomaspoignant:
# Description
Create a helm for the relay-proxy to make it easy to deploy it inside Kubernetes.
This is a basic **helm chart**, it put the `relay-proxy` configuration as `configmap` and mount it as a volume in the pod.

## What is missing before being able to merge it
- [x] update `appVersion` in the `Chart.yaml` file when we release a new version.
- [x] release the chart using [helm/chart-releaser](https://github.com/helm/chart-releaser)
- [x] lint the chart during CI if something changes in this folder.
- [x] update the documentation to mention this chart
 
# Changes include

# Closes issue(s)
Resolve #530 

# Checklist
- [x] I have tested this code


cmd/relayproxy/helm-charts/relay-proxy/goff-proxy.yml
```diff
@@ -0,0 +1,9 @@
+---
+listen: 1031
+pollingInterval: 1000
+startWithRetrieverError: false
+retriever:
+  kind: http
```
Rongtao-Fu-Imprivata, line 6: We plan to use GitHub retriever because we have standardized on GitHub as the single source repository, which use githubToken parameter to pass the token. We plan to use K8s secret to pass the token for obvious security reason, and need to replace the githubToken value with the provided K8s secret. It will be great to illustrate how to use K8s secret to provide credentials in the relay proxy config file, which is stored as configmap and mounted as a volume by this Helm chart.

Rongtao-Fu-Imprivata: I also asked our DevOps engineers to review this PR tomorrow, and see if they have additional comments.

lorelei-rupp-imprivata:
I work with RongTao above - and am on the DevOps team. We took a look at this chart, thanks for such a fast response on our gitissue by the way!
We had some questions on this that I wanted to leave here:

1. First up on the configuration of the relay proxy (https://gofeatureflag.org/docs/relay_proxy/configure_relay_proxy)
Having this be required in a file in your helm chart doesn't really scale for public adopters of this, if this helm chart were in a public helm chart repo. Other helm patterns I have seen in many various K8 projects usually allow for the config to be piped through in the values file, something like this :
```
values.yaml
  relay.config: |
     config1: thing
     config2: thing
     config3: thing
```
Then that could be in turn piped into the config map if you wanted. This would be more optimal over a local file
Also NO sensitive data or secret configuration should be supported or allowed here either
We would like to see perhaps another section in the values.yaml for any relay configuration that is a "secret". One example is the `githubtoken` this is a secret and cannot be in a yaml file, or in values file in plain text or even encrypted in the values file. This needs to be in a kubernetes secret in the cluster and read in to your docker container.

2. A nice feature enhancement on the relay proxy would be to support ENV var overrides of the relay proxy configuration. This would allow for easily injecting the pieces of the configuration that are "secret" too as well as overwriting any default configuration that might be set in the values.yaml as suggested in 1. above.

3. The relay configuration for the githubtoken - it says it needs "repo" permission on the PAT, which gives full private repository access, we wanted to confirm what sort of access the relay proxy really needs, do you really just need "read" but github PATs dont support that level?

4. Can we get resource requests and limits added to the helm chart for the container?

thomaspoignant:
Hey, @lorelei-rupp-imprivata, @Rongtao-Fu-Imprivata, thanks a lot for the details reviews.
Let me try to answer you in detail too.

1. I agree we could put the configuration directly in the `Values.yaml` and if we still want to use a configuration file we could use `--set-file key=filepath`.
```
 --set-file stringArray                       set values from respective files specified via the command line (can specify multiple or separate values with commas: key1=path1,key2=path2)
```
What do you think about this option?
It will also enable the possibility to override a dedicated key by overriding it, which will solve your issue with secrets.

2. Super idea, since we are using `viper` it should be easy to do it.
I have opened an issue for that: #532

3. Yes we need only read-only access for the github retriever. 

4. You can add them yourself for sure, but it is hard for me to provide these resource limits for you, it really depends on your usage.

lorelei-rupp-imprivata:
1. - Yeah I think allowing for config file path is fine too. While you could use the secrets this way too with the set commands, normally this is fully automated for us, so we are not using cli helm commands in this manner. We would use terraform or ArgoCD to roll this helm chart out, and we would also roll out an External Secret with the token in it, and that would pull it from AWS Secrets Manager, and then create the Kuberntes secret. What we would want to pass to the helm chart would be the existing kubernetes secret name, or be able to use ENV vars directly to inject into the container the secret
2. Looks like you already implemented this?!?! Thanks! Ideally the helm chart could support a deployment section with an env section then you could just inject directly into the container, i have seen this pattern a lot on many helm charts
4. So I was more interested in requests not the limits. Yes limits would be good to be set by us or your users, but the requests should be set by this project I would assume. but exposing both of these on the deployment spec and in the values would be great for configuration

thomaspoignant:
@lorelei-rupp-imprivata I have updated the chart (+ the README) to be able to set the `relay-proxy` configuration directly in the `Values.yaml` file.

Can you check this commit https://github.com/thomaspoignant/go-feature-flag/pull/531/commits/5649a971724219448c67215c3f56ba43d6e803be and tell me if it is what you had in mind?

lorelei-rupp-imprivata:
> @lorelei-rupp-imprivata I have updated the chart (+ the README) to be able to set the `relay-proxy` configuration directly in the `Values.yaml` file.
> 
> Can you check this commit [5649a97](https://github.com/thomaspoignant/go-feature-flag/commit/5649a971724219448c67215c3f56ba43d6e803be) and tell me if it is what you had in mind?

yeah that totally works. I have also seen people do it as one giant string too, but either is fine!

thomaspoignant:
@lorelei-rupp-imprivata @Rongtao-Fu-Imprivata I have released the helm chart with version `v1.4.0` of **GO Feature Flag**.

You can check this [documentation](https://gofeatureflag.org/docs/relay_proxy/deploy_relay_proxy#deploy-in-kubernetes-using-helm) to see how to use it.
I have deployed a helm repo available at https://charts.gofeatureflag.org to download it directly.

You can override the `values.yaml` as you wish.
Let me know if you need more information, and thanks a lot for the feedback.

PR #649, opened 2023-03-30
relay-proxy: Support API Keys validation to limit the applications being able to use GO Feature Flag

### LINKED ISSUE #613: relay-proxy: Support API Keys validation to limit the applications being able to use GO Feature Flag
# Proposal
The proposed solution is to modify the relay proxy to support a list of authorized API keys in the configuration. Authorized sources and providers can obtain an API key by contacting the administrator of the relay proxy. The administrator will add the API key to the list of authorized API keys in the configuration, and the associated providers can then use the API key in their requests to the relay proxy.

To use the API key, providers will include it as an `Authorization` header in their HTTP request. The relay proxy will check the API key in the `Authorization` header against the list of authorized API keys in the configuration. If the API key is valid, the request will be processed, and the response will be returned. If the API key is not valid, the request will be rejected with an appropriate error message.

# Requirements
The feature should meet the following requirements:
- The relay proxy should be modified to support a list of authorized API keys in the configuration.
- The authorized API keys list should be configurable.
- The relay proxy should check the API keys in the Authorization header when receiving any requests and only allow requests with valid API keys.
- Unauthorized requests should be rejected with an appropriate error message.
- The feature should not significantly impact the performance of the relay proxy.

# User Interactions
To add a new provider and obtain an API key, you will have to contact the administrator of the relay proxy. This administrator can configure as many API Keys as needed.

Applications can then include the API key in the provider configuration and it will be sent as an Authorization header in the HTTP request to the relay proxy. 
- If the API key is valid, the request will be processed, and the response will be returned. 
- If the API key is not valid, the request will be rejected with an error message indicating that the API key is invalid.

Monitoring endpoints should not be protected by any API Key (`/info`, `/health`, `/metrics`)

# Acceptance Criteria
The feature will be considered complete when:
- The relay proxy has been modified to support a list of authorized API keys in the configuration.
- The authorized API keys list can be configured and updated by the administrator of the relay proxy.
- The relay proxy checks the API keys in the Authorization header when receiving any requests and only allows requests with valid API keys.
- Unauthorized requests are rejected with an appropriate error message.
- Providers are ready to use the API Keys.

dhanusaputra:
# Description

> The feature should meet the following requirements:
>    * The relay proxy should be modified to support a list of authorized API keys in the configuration.

will use `AdminAPIKeys` to initiate list of admin keys when proxy first start

>   * The authorized API keys list should be configurable.

configurable via apiKeyGet, apiKeyDelete, apiKeyCreate

>    * The relay proxy should check the API keys in the Authorization header when receiving any requests and only allow requests with valid API keys.

use `middleware.KeyAuthWithConfig` to check the header with format: `Bearer <key>`

>    * Unauthorized requests should be rejected with an appropriate error message.

`middleware.KeyAuthWithConfig.Validator` will reject unauthorized requests

>    * The feature should not significantly impact the performance of the relay proxy.

basically will add one more process in middleware to check the header and use internal memory for storing api keys

# Changes include

# Closes issue(s)

Resolve #613 

# Checklist
- [ ] I have tested this code


dhanusaputra:
hi @thomaspoignant just drafted this PR for initial feedback
please help take a look when have time thank you

please also check the controller part cmd/relayproxy/controller/api_key.go
it's still WIP, have put some comments regarding the API contract flow

one more thing, currently we can put `user` in request body of some endpoints to know who is the requester
maybe we can use the user data when we create API key?

thomaspoignant (commented):
Hey thanks for this PR.

I am not sure to understand how you proposal will work.
How can I add an API key? By calling the API and it will be stored in memory?
How will it work when the relay-proxy will restart, will we lose the API keys configured?

How I was thinking of doing this, was to have a list of valid API keys in the relay proxy configuration file and I would have challenged the authorization headers with everything that is part of this list.

dhanusaputra:
> How can I add an API key? 

using the `apiKeyCreate` endpoint

> By calling the API and it will be stored in memory? 

yes

> How will it work when the relay-proxy will restart, will we lose the API keys configured?

yes everything will be lost

> How I was thinking of doing this, was to have a list of API keys in the relay proxy configuration file and I would have challenged the authorization headers with everything that is part of this list.

got it, let me rework the PR, I'm more aligned this way too, thank you for your feedback!

cmd/relayproxy/README.md
```diff
@@ -47,7 +47,7 @@ Before starting your **relay proxy** you will need to create a minimal configura
 # this is a minimal config containing only where your flag file is located 
 retriever:
   kind: http
-  url: https://raw.githubusercontent.com/thomaspoignant/go-feature-flag/main/examples/file/flags.yaml
```
dhanusaputra, line 50: invalid url

thomaspoignant: Good catch 👍

cmd/relayproxy/config/config.go
```diff
@@ -137,6 +138,9 @@ type Config struct {
 
 	// Version is the version of the relay-proxy
 	Version string
+
+	// APIKeys list of API keys that authorized to use endpoints
+	APIKeys map[string]bool `mapstructure:"apiKeys"`
```
thomaspoignant, line 143: I am not sure to understand why you have a `map[string]bool`, is it to have a set here?
If so we may need to consider having a `map[string]interface{}` instead since it is a bit better in term of memory and size.

cmd/relayproxy/testdata/config/valid-file.yaml
```diff
@@ -7,3 +7,6 @@ retriever:
 exporter:
   kind: log
 enableSwagger: true
+apiKeys:
+  apikey1: true
+  apikey2: true
```
thomaspoignant, line 12: I am not sure to understand the boolean here.
Would it be better to have only a list, something like 
```yaml
apiKeys:
  - apikey1 # with a comment explaining who is the owner of the key
  - apikey2
```

I think it will make the configuration representation way easier to understand.
You can still put it in a set in the config struct, but in terms of configuration, it will be way more understandable for the users of the relay proxy.

website/docs/relay_proxy/configure_relay_proxy.md
```diff
@@ -31,6 +31,7 @@ If you want to replace a nested fields, please use `_` to separate each field _(
...
 | `exporter`                | [exporter](#exporter)     | **none**    | Exporter is the configuration on how to export data.                                                                                                                                                                                                                                          |
 | `notifier`                | [notifier](#notifier)     | **none**    | Notifiers is the configuration on where to notify a flag change.                                                                                                                                                                                                                              |
+| `apiKeys` | map[string]bool | **none** | List of authorized API keys. Each request will need to provide one of authorized key inside `Authorization` header with format `Bearer <api-key>`. There will be no authorization when this config is not configured.<br /><br />_Note: there will be no authorization when this config is not set._ |
```
thomaspoignant, line 34: You mention twice this, maybe we should remove one.
> There will be no authorization when this config is not configured.

thomaspoignant (changes requested):
This approach is way closer to what I had in mind thank you 🙏 
I've added a few review point because I am not sure to get all the points here.

cmd/relayproxy/config/config.go
dhanusaputra: thanks for the article, TIL
let me change it into slice since it's more easy to understand in config

website/docs/relay_proxy/configure_relay_proxy.md
dhanusaputra: these 2 docs looks similar
https://github.com/thomaspoignant/go-feature-flag/blob/HEAD/cmd/relayproxy/docs/configuration.md
https://github.com/thomaspoignant/go-feature-flag/blob/main/website/docs/relay_proxy/configure_relay_proxy.md
maybe can just keep one

thomaspoignant: You are right.
I guess we could delete the one in `cmd/relayproxy/docs` since it is not really used and keep only the one that finish on the website.

thomaspoignant (commented):
This looks exactly to what I had in mind.
I will review it soon.

Issue #662, opened 2023-04-06
(feature) Implement Open-feature web provider.


thomaspoignant:
### Requirements

Now that the Open Feature [web SDK](https://github.com/open-feature/js-sdk/tree/main/packages/client) is available it could be great to implement a client provider for GO Feature Flag for the web SDK.

To continue to support community initiatives we will place the provider in the [js-sdk-contrib](https://github.com/open-feature/js-sdk-contrib/tree/main/libs/providers) repository.

Please create a new folder called `go-feature-flag-web` and ping @thomaspoignant for review in the pull request on the contrib repo.

thomaspoignant:
A first version of the provider is in review in the open-feature contrib repository: https://github.com/open-feature/js-sdk-contrib/pull/474

PR #802, opened 2023-05-17
Websocket notifier implementation in the relay proxy

thomaspoignant:
# Description
Implement in the relay proxy a websocket component that notifies when a flag is changing to allow the providers to refresh their internal cache.

# Changes include

# Closes issue(s)
We need this for #662 to be able to implement it.

# Checklist
- [x] I have tested this code


thomaspoignant:
> @thomaspoignant is this about sending realtime flag update to backend services which depends on SDKs? Just curious.

Yes and no, the goal for now is for the front end SDK to know that they have to refresh their configuration, but when this will be in place we can also have the realtime flag update with a similar system.

thomaspoignant (commented):
Note to my self, we should work on the TODO.

PR #989, opened 2023-08-10
Add a shutdown task after tests

thomaspoignant:
Add a shutdown task after tests


thomaspoignant:
This is blocked because of this issue https://github.com/open-feature/java-sdk/issues/552

PR #1105, opened 2023-09-25
Feature/add event source field

### LINKED ISSUE #978: (feature) Adding event source field
### Introduction
Evaluation event is used for statistics purposes.  
The event can be published by the proxy, or from the [provider](https://github.com/open-feature/java-sdk-contrib/pull/369).

### Feature
Enrich event by adding it _source_ field which represents the source which publishing the message:
- PROXY: set by the proxy when evaluated
- PROVIDER_CACHE: set by the provider when evaluated from cache

### Tasks
- server-side: Support it at server-side for receiving the optional field and enrich the statistics by it.
- Proxy: set PROXY source by the proxy when evaluated
- OpenFeature Providers: set PROVIDER_CACHE source by the provider when evaluated from cache.
  This is for providers which implemented cache, currently GO and Java.

andrebenedetti:
# Description

Added `Source` field on  FeatureEvent

Source is set to "PROVIDER_CACHE" for events received through the CollectEvalData endpoint. Otherwise, it is set to "SERVER"

# Changes include

# Closes issue(s)

Resolve #978

# Checklist
- [x] I have tested this code


andrebenedetti:
hey, @thomaspoignant, please have a look whenever you have some time =) 

I failed to see my changes to the docs locally :thinking: , also I'm taking a look at the failed pipeline steps

thomaspoignant:
I haven’t check the PR yet but if you want to have swagger CI to pass you will have to generate the new swagger schema.

To do this please run `make swagger` and commit the files.

cmd/relayproxy/controller/collect_eval_data.go
```diff
@@ -47,6 +47,7 @@ func (h *collectEvalData) Handler(c echo.Context) error {
 	}
 
 	for _, event := range reqBody.Events {
+		event.Source = "PROVIDER_CACHE"
```
thomaspoignant, line 50: I think we should not override the `Source` if it is already provided in the `event`.
What do you think if we have something like this instead:
```suggestion
		if event.Source == "" {		
		    event.Source = "PROVIDER_CACHE"
		}
```

exporter/logsexporter/exporter.go
```diff
@@ -33,7 +33,7 @@ type Exporter struct {
 // Export is saving a collection of events in a file.
 func (f *Exporter) Export(_ context.Context, logger *log.Logger, featureEvents []exporter.FeatureEvent) error {
 	f.initTemplates.Do(func() {
-		// Remove bellow after deprecation of Format
+		// Remove below after deprecation of Format
```
thomaspoignant, line 36: good catch 👍

thomaspoignant (changes requested):
Thanks a lot, @andrebenedetti this is a great addition to GO Feature Flag.

I've only 1 small feedback to be sure we are not overriding the field if it is already set in the event. We should also probably add a test for this use-case.

andrebenedetti:
@thomaspoignant good points

I'll make those changes later today. Thanks for reviewing

cmd/relayproxy/controller/collect_eval_data.go
andrebenedetti: Do we also want to check that when the source is SERVER? Trying to think of an use case in which that could be useful 🤔

thomaspoignant: Yes good point.
Be awesome if you could do it.

i can help with that if you want.

andrebenedetti: this one seems a little bit trickier

well, if you have time, help is always welcome :slightly_smiling_face:, but don't worry if you don't, I can give it a try

thomaspoignant: I was thinking about it again this morning, and I think this use case is not really possible (overriding when in `SERVER` mode).

thomaspoignant (approved):

thomaspoignant:
Thanks for the awesome work on this PR.
I will merge it and it will be available in the next version of GO Feature Flag.

andrebenedetti:
> Thanks for the awesome work on this PR. I will merge it and it will be available in the next version of GO Feature Flag.

🎆 thanks for the support
It was nice to learn more about the repo
We use LaunchDarkly at work and, but go-feature-flag would cover all our needs if we were to switch :). I'll let the team know about it and maybe one day we'll migrate if it aligns with the business goals

thomaspoignant:
Oh yes, let me know.
Launchdarkly is way more complete, but sometimes it is too much (and also expensive 😉).

In any case I would be happy to get your feedback about the repo.

PR #1285, opened 2023-11-20
Support of evaluation context enrichment

### LINKED ISSUE #1270: (feature) Add ability for proxy to inject context
### Requirements

I'd like to be able to add additional context at the proxy level that can be used in addition to the context provided by the client through the proxy.

For example, the user requests a feature flag passing in email address. The proxy adds additional context that was configured by it, such as the environment in which the proxy is run, such as 'production: true/false'. These two values are then used by the rules to return the desired flag. 

The use cases are wide-ranging, but essentially consist of context that a generic service or client using the proxy does not know or need to know, but affects the rule processing. This tends to be very necessary when you have multiple deployed environments, where there is a set of dev, staging or production environments and you want differing behaviour between them but don't want to pollute services with environment-specific context.

It may be that this feature exists, but I was unable to find a way to do this.

I had a quick look at the proxy code, and it seems that additional configuration in the form of a context section would be required and then the values in there would be merged with with the incoming client context (probably with the proxy context taking precedence for any matching keys), before passing down to the go-feature module itself.

thomaspoignant:
# Description
The common evaluation context is a new mechanism that enriches all the evaluation contexts with a common context of the execution server.

This is particularly useful when you support different environment variables, or if you want to target a specific server.

To use it, you can configure the field `EvaluationContextEnrichment` (`map[string]interface{}`), so you can add everything you want to enrich the request.

:warning: In case we have the same key in the `EvaluationContextEnrichment` and in the `EvaluationContext`, the `EvaluationContextEnrichment` will override what was set in the `EvaluationContext`.

## Example
If you configure your `EvaluationContextEnrichment` like this:
```
"serverName": "server01"
"IP": "8.8.8.8"
"environment": "prod"
```

And you evaluation your flag with this context:
```
"name": "John Doe"
"email": "john.doe@gofeatureflag.org"
"key": "123456789"
```

Your context will be merged and you will be able to create rules about all those fields:
```
"serverName": "server01"
"IP": "8.8.8.8"
"environment": "prod"
"name": "John Doe"
"email": "john.doe@gofeatureflag.org"
"key": "123456789"
```

# Changes include

# Closes issue(s)
Resolve #1270 

# Checklist
- [x] I have tested this code


richardj-bsquare (approved):
Looks like the new evaluation context should meet the requirements outlined in #1270

Looking forward to trying it out!

PR #1407, opened 2023-12-26
✨  Feature enhance rule editor

### LINKED ISSUE #1372: (feature) editor - Create a better rule editor
## Requirements

In the flag editor _(https://editor.gofeatureflag.org)_ creating a rule is a simple textarea.
It would be great to have a more advanced rule editor supporting making it easy the creation of a complex rule.

Supported case are the following https://gofeatureflag.org/docs/configure_flag/rule_format#query-format

One idea would be to have something like this:
Image

But I am happy to challenge this if someone has a better idea on how to have a great UX for that.

The source of the editor are available https://github.com/thomaspoignant/go-feature-flag/tree/main/website/src/pages/editor

Esequiel378:
# Description
Enhance the rules editor by updating the UI to use a query editor. The query editor allows users to create complex rules using groups and leverage drag and drop to be able to tweak the rule as needed.

# Changes include

# Closes issue(s)
Resolve #1372 

# Checklist
- [x] I have tested this code

image

~[image](https://github.com/thomaspoignant/go-feature-flag/assets/40343304/e7bc852a-53c4-49c7-8aaf-c5ca4189c35c)~

~[previous image](https://github.com/thomaspoignant/go-feature-flag/assets/40343304/242aa7ef-c827-424f-8f15-ca0cfa0615e0)~

~[previous image](https://github.com/thomaspoignant/go-feature-flag/assets/40343304/4eada2c7-094c-45c2-9df8-ad6cf2b83851)~


thomaspoignant:
Hey @Esequiel378 this PR looks super promising 👍 
Let me know when you want a review and I will have a closer look, but I like what I see [here](https://deploy-preview-1407--go-feature-flag-doc-preview.netlify.app/editor)

Esequiel378:
Hi @thomaspoignant 👋 ! The branch is ready for review, but I found that after running the `eslint:fix` script, a lot of files got updated (files I didn't work on btw), what do you think I should do with them? I can iterate over each file and revert those formatting changes  if you like

website/src/components/editor/Rule/index.js
```diff
@@ -109,3 +138,127 @@ export function Rule({variations, label, isDefaultRule}) {
...
+  return json.rules.map(processRule).join(` ${json.combinator} `);
+}
+
+const ruleOperators = [
+  {name: '==', label: 'Equals To'},
+  {name: '!=', label: 'Not Equals To'},
+  {name: '<', label: 'Less Than'},
+  {name: '>', label: 'Greater Than'},
+  {name: '<=', label: 'Less Than Equal To'},
+  {name: '>=', label: 'Greater Than Equal To'},
```
thomaspoignant, line 257: I am more in favour of using the text notation everywhere.
```suggestion
  {name: 'eq', label: 'Equals To'},
  {name: 'ne', label: 'Not Equals To'},
  {name: 'lt, label: 'Less Than'},
  {name: 'gt', label: 'Greater Than'},
  {name: 'le', label: 'Less Than Equal To'},
  {name: 'ge', label: 'Greater Than Equal To'},
```

thomaspoignant (commented):
I had a first look at the PR and I like it.
My only doubt is a styling thing, I am wondering if we can use the same components for the buttons _(not sure if the library allows it?)_, select and inputs as the rest of the page?
The main reason behind that is because we don't have any information in the inputs on what we are editing 

With the actual style we don't really know where to put the key name and where to put the value.
I would suggest to have `key` and `value` as field name here, WDYT?
Screenshot 2024-01-03 at 21 19 05

Despite that, I really like the drag and drop in it this is super impressive.

_And thanks for fixing the eslint issues_ 🙏

Esequiel378:
@thomaspoignant I've updated the UI based on your last comment and it looks great! Thanks for the suggestion 🤗.
I just realized that I'm missing the query validation, since is pretty easy to end up with an invalid query. I will keep you posted on this topic

website/src/components/editor/Rule/index.js
```diff
@@ -1,30 +1,49 @@
...
 Rule.propTypes = {
   variations: PropTypes.array,
   label: PropTypes.string.isRequired,
   isDefaultRule: PropTypes.bool,
 };
 export function Rule({variations, label, isDefaultRule}) {
-  const {register, watch} = useFormContext();
+  const [query, setQuery] = useState({
+    combinator: 'and',
+    rules: [{field: '', operator: '==', value: ''}],
```
thomaspoignant, line 29:
```suggestion
    rules: [{field: '', operator: 'eq', value: ''}],
```

Just to keep consistency with what we have in the `Select`.

thomaspoignant (commented):
Hey @Esequiel378 this looks way better I love it <3 
I have some quick remarks still to be sure it works well.

- When we are comparing something to a string we should put the string between `"`, for now it is put as free text and it will break the rule.
- I will also try not to add the `and`/`or` if the fields are blank _(put this is a nice to have)_.

Despite those 2 things I think we are close to merge this PR 🙌

website/src/components/editor/Rule/index.js
Esequiel378: Sure thing! Just to clarify this is the way the query should look like?

- `env eq "test"`
- `env eq "prev" and user_id in [1, 2, 3, 4]`

Also, since I don't have a way to know whether a field should be a string or a number, should I always treat values as strings? For example, the above list would be converted like this `["1", "2", "3"]`

thomaspoignant: I guess we can try to do this based on the type of content.
But you can probably put everything as string for now if it is easier, and we can evolve it later.

thomaspoignant (approved):
@Esequiel378 Thanks a lot for your contribution this is awesome and make the editor way better.

I'll merge this PR and deploy the editor 🚀

PR #1422, opened 2024-01-02
[BUGFIX] Helm conditional for HPA K8s API version

lrvan:
# Description
`autoscaling/v2beta1` is deprecated in K8s v1.25 and should be replaced with `autoscaling/v2` since K8s v1.23. This change adds Helm conditional to check whether K8s API supports newer version and use it in the rendered template if possible.

# Changes include

# Checklist
- [x] I have tested this code


thomaspoignant (approved):
Good catch @lrvan I will deploy a new version with the fix today.

PR #1426, opened 2024-01-02
Feature: Add Kafka exporter

### LINKED ISSUE #89: (feature) Add a kafka data exporter
# Requirements
Add a kafka data exporter but this time since we are using an async system we should produce an event everytime someone is calling a variation.

ccuetoh:
# Description
As requested by #89 I'm proposing code that implements a Kafka exporter. This code uses IBM's [sarama](https://github.com/IBM/sarama) library as the base client for Kafka.

# Changes include

# Closes issue(s)
Resolve #89

# Checklist
- [x] I have tested this code


ccuetoh:
Hi @thomaspoignant, happy new year! This is a (very) rough first draft for the Kafka exporter (still missing tests, docs, more testing, etc.). I'm checking to see if I'm going in the right direction before continuing.

In particular, I'm running into two questions:
- Configuration: I'm trying to avoid polluting _too_ much the exporter configuration struct. I've added a custom config definition for Kafka, but I see that's not how you've done it elsewhere. Can I have your input on this?

- Async vs Sync: Kafka is designed to work in either synchronous or asynchronous mode, and the Kafka library I've chosen supports both. My understanding is that by design the exporters will always be called asynchronously, and thus the message publishing should be synchronous as doing it asynchronously would be redundant. Is this assumption correct?

exporter/kafkaexporter/exporter.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return fmt.Errorf("send: %w", err)
+	}
+
+	fflog.Printf(logger, "info: [KafkaExporter] sent %d messages", len(messages))
+	return nil
+}
+
+func (f *Exporter) IsBulk() bool {
+	return true
```
thomaspoignant, line 70: To answer your question 
>Async vs Sync: Kafka is designed to work in either synchronous or asynchronous mode, and the Kafka library I've chosen supports both. My understanding is that by design the exporters will always be called asynchronously, and thus the message publishing should be synchronous as doing it asynchronously would be redundant. Is this assumption correct?

If you do this change

```suggestion
	return false
```

The exporter will not be async anymore and will send every event 1 by 1 to the exporter.
This is what we are doing also in the SQS exporter _([see here](https://github.com/thomaspoignant/go-feature-flag/blob/fd90988686cccb85f23cdedc39f7e2bc65c47380/exporter/sqsexporter/exporter.go#L74))_.

Since Kafka is designed to receive the events 1 by 1, I think we need to set this to false.

exporter/kafkaexporter/exporter.go
```diff
@@ -0,0 +1,102 @@
...
+	// Default: JSON
+	Format string
+
+	// CsvTemplate is used if your output format is CSV.
+	// This field will be ignored if you are using another format than CSV.
+	// You can decide which fields you want in your CSV line with a go-template syntax,
+	// please check internal/exporter/feature_event.go to see what are the fields available.
+	// Default:
+	// {{ .Kind}};{{ .ContextKind}};{{ .UserKey}};{{ .CreationDate}};{{ .Key}};{{ .Variation}};{{ .Value}};{{ .Default}}\n
+	CSVTemplateString string
```
thomaspoignant, line 28: I am not sure if I would accept CSV to push the message in Kafka.
I would start by pushing in JSON only, and if we want to have different formats I would for `avro` which is probably more common for the kafka users.

thomaspoignant (commented):
Hi @ccuetoh, thanks a lot for this contribution this is super great to see this coming 🙌 
and Happy New Year also 🎉.

Let me try to answer your questions:

1. I don't have strong feelings about putting the configuration in a different struct, but I will do that only if the number of parameters is big. In your case, I would keep it in the exporter directly, especially because some others are already inside _(`Format`, ...)_.
2. I have answered this one directly in the code, but if you change the bulk option the exporter becomes synchronous and I think this is what we should do here.

Except that, I am really happy that you are using IBM/sarama this is a really good choice here 👍

ccuetoh:
Mostly done. I need to update the documentation and take a look at that linter issue 😬

thomaspoignant:
> Mostly done. I need to update the documentation and take a look at that linter issue 😬

Awesome @ccuetoh, let me know when you will need a full review.

ccuetoh:
@thomaspoignant, I'm ready for a full review

exporter/kafkaexporter/exporter.go
```diff
@@ -0,0 +1,122 @@
...
+
+// Export will produce a message to the Kafka topic. The message's value will contain the event encoded in the
+// selected format. Messages are publish synchronously and will error immediately on failure.
+func (e *Exporter) Export(_ context.Context, logger *log.Logger, featureEvents []exporter.FeatureEvent) error {
+	if e.sender == nil {
+		err := e.initializeProducer()
+		if err != nil {
+			return fmt.Errorf("writer: %w", err)
+		}
+	}
```
thomaspoignant, line 57: I had an issue with the initialization here.
During my test what I've done is try to configure the exporter like this: 

```go
DataExporter: ffclient.DataExporter{
  FlushInterval:    1 * time.Second,
  MaxEventInMemory: 100,
  Exporter: &kafkaexporter.Exporter{
    Settings: kafkaexporter.Settings{},
  },
},
```

And what happened:
- The first call to the Kafka exporter went well because the `e.initializeProducer()` return an error
- But the second call `panic` because of the `e.init.Do` it is not throwing any error and we try to send a message with `e.sender.SendMessages(messages)` when `e.sender` is `nil`.

exporter/kafkaexporter/exporter.go
```diff
@@ -0,0 +1,122 @@
...
+
+	init   sync.Once
+	sender MessageSender
+	// dialer will create the producer. This field is added for dependency injection during testing as sarama
+	// has the annoying tendency to dial as soon as a producer is created.
+	dialer func(addrs []string, config *sarama.Config) (MessageSender, error)
+}
+
+// Export will produce a message to the Kafka topic. The message's value will contain the event encoded in the
+// selected format. Messages are publish synchronously and will error immediately on failure.
```
thomaspoignant, line 50:
```suggestion
// selected format. Messages are published synchronously and will error immediately on failure.
```

cmd/relayproxy/config/exporter.go
```diff
@@ -59,6 +61,7 @@ const (
 	S3Exporter            ExporterKind = "s3"
 	GoogleStorageExporter ExporterKind = "googleStorage"
 	SQSExporter           ExporterKind = "sqs"
+	KafkaExporter         ExporterKind = "kafka"
```
thomaspoignant, line 64: For now your exporter is not working in the relayproxy because `func (r ExporterKind) IsValid() error` is returning an error.

Be sure to add `KafkaExporter` in the list of the valid `ExporterKind`.

cmd/relayproxy/service/gofeatureflag.go
```diff
@@ -159,92 +162,118 @@ func initRetriever(c *config.RetrieverConf) (retriever.Retriever, error) {
...
 
+		return &sqsexporter.Exporter{
+			QueueURL:  c.QueueURL,
+			AwsConfig: &awsConfig,
+		}, nil
+	case config.KafkaExporter:
+		return &kafkaexporter.Exporter{
+			Format:   format,
+			Settings: c.Kafka,
+		}, nil
```
thomaspoignant, line 274: During my tests I've seen that the `Settiings` key has always empty value for the fields.
It may be because there are no annotations for the fields in the retriever? 

This is the configuration file I use for the relayproxy:
```yaml
listen: 1031
pollingInterval: 1000
retriever:
  kind: file
  path: ./openfeature/provider_tests/flags.yaml
exporter:
  kind: kafka
  format: json
  settings:
    topic: go-feature-flag-events
    addresses:
      - localhost:29092
enableSwagger: true
```

thomaspoignant (changes requested):
Hey @ccuetoh, thanks a lot for this contribution.
I've checked your PR and found some issues.

1. There is something not working well during the initialization if an error happen _(see details in the PR)._ It works well during the 1st call be the second one is failling.
2. The relay proxy integration is not working currently, the `kind: Kafka` is not consider valid and the parameters are not passed to the configuration object.
3. I think there is a missing documentation for the relay proxy configuration.

_PS: I've added an example for this `exporter`, this is what I used to test the lib version and I found it cool to have it in the list of the example._

exporter/kafkaexporter/exporter.go
ccuetoh: I've replicated the issue now, my testing setup did not catch this previously, sorry. What would be the preferred behavior for calling the exporter after a failure? Should I continue to retry the connection?

cmd/relayproxy/service/gofeatureflag.go
ccuetoh: The configuration struct expects the `kafka` key for the settings. I'm not 100% sure about the name though, since it might be a bit repeitive. Using `settings` is too generic too. Maybe `kafka_settings`?

website/docs/relay_proxy/configure_relay_proxy.md
```diff
@@ -214,6 +215,15 @@ _To understand the format in which a flag needs to be configured in MongoDB, che
...
+|------------------|----------|-------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `kafka.topic`    | string   | **none**          | **(mandatory)** Kafka topic to bind to.                                                                                                                                                                                                                                                                    |
+| `kafka.adresses` | []string | **none**          | **(mandatory)** List of bootstrap addresses for the Kafka cluster.                                                                                                                                                                                                                                         |
```
thomaspoignant, line 223:
```suggestion
| `kafka.addresses` | []string | **none**          | **(mandatory)** List of bootstrap addresses for the Kafka cluster.                                                                                                                                                                                                                                         |
```

thomaspoignant (approved):
@ccuetoh thanks again for this contribution this is super great and it was something I was wanting since a long time 👍 

I will merge it and release it in the next version.

PR #1449, opened 2024-01-15
(feature) Passing the Evaluated Rule Name to the Metadata

carlossilveirajunior:
# Description

The Evaluated Rule Name is needed during our experiment analysis phase. During this phase, we need to keep track of the user that has been assigned to a defined variant. Our experiment tends to have different rules, such as internal experiments, free account experiments, etc... Those phases can happen in parallel, and we need to define how the phase is performing. That way the $reason$ on the response is not enough and the metadata must be enhanced. 

Because it is possible to set a variant to a target group, if we don't keep track of the rule, we can have a Sample Mismatch. E.g. if we set enterprise user to variant off, and not enterprise user to 50:50 on and off. That means we will have many more off-users than on. In this case, the reason is enough, but we have the case where enterprise can also be split on a small percentage causing the mismatching.

# Changes include
- [X] New feature (non-breaking change that adds functionality)

# Closes issue(s)

Resolve #

# Checklist
- [X] I have tested this code

### DIFF
--- a/cmd/editor/main.go
+++ b/cmd/editor/main.go
@@ -59,21 +59,9 @@ func EvaluateHandler(c echo.Context) error {
 		ErrorCode:     resolutionDetails.ErrorCode,
 		Failed:        resolutionDetails.ErrorCode != "",
 		Cacheable:     resolutionDetails.Cacheable,
-		Metadata:      constructMetadata(f, resolutionDetails),
+		Metadata:      f.GetMetadata(),
 	}
 	return c.JSON(http.StatusOK, resp)
-}
-
-func constructMetadata(f flag.InternalFlag, resolutionDetails flag.ResolutionDetails) map[string]interface{} {
-	metadata := f.GetMetadata()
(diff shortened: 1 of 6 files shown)


thomaspoignant (changes requested):
Hey @carlossilveirajunior this is a good idea to had the rule name in the metadata's.

Your PR is great but I would recommend to add some tests and documentation before we can merge this and add it to the product.

carlossilveirajunior:
> Hey @carlossilveirajunior this is a good idea to had the rule name in the metadata's.
> 
> Your PR is great but I would recommend to add some tests and documentation before we can merge this and add it to the product.

Hey @thomaspoignant, first of all, thanks for the fast reply. I have added a test, and doc on the function I created. If we see there is a place to create other docs, please let me know.

That is my first PR on your repo, any other feedback is really welcome. And I hope that soon we can push it to the product to unblock us from using it. We, on Miro.com, are very excited to start using Go-Feature-Flag!!

thomaspoignant (approved):
Thanks again for your contribution, it will be released in the next version.

PR #1548, opened 2024-02-12
Ofrep

thomaspoignant:
This PR will stay open for a while.

This is an implementation of the `OpenFeature Remote Evaluation` Protocol which is under the specification phase.
The goal is to be able to support OFREP as soon as it is available.


thomaspoignant:
This PR is missing test but OFREP is an experimental feature, we should not forget to add them later when OFREP is more stable.

PR #1605, opened 2024-02-28
Fix format of the Python example of OpenFeature in README

ChihweiLHBird:
Just a simple doc fix


thomaspoignant:
Oh good catch @ChihweiLHBird.
Thanks a lot for your contribution.

thomaspoignant (approved):

PR #1663, opened 2024-03-20
fix panic concurrent map writes when constructing metadata

### LINKED ISSUE #1662: (bug) Panic when call getVariation with rules evaluated in parallel
### Observed behaviour

We've encountered panic when http requests come in parallel and retrieving flag at the same time.

image

### Expected Behavior

Retrieve flag usecase should be concurrent safe

### Steps to reproduce

Just make a test and run it in parallel. Or you also can reproduce it via http load test tools with the endpoint that being hit should call any function that try to retrieve flag.

image

image

kbiits:
# Description

# Changes include

# Closes issue(s)

Resolve #1662 

# Checklist
- [x] I have tested this code

### DIFF
--- a/variation.go
+++ b/variation.go
@@ -415,7 +415,7 @@ func getVariation[T model.JSONType](
 
 	flagCtx := flag.Context{
 		DefaultSdkValue:             sdkDefaultValue,
-		EvaluationContextEnrichment: g.config.EvaluationContextEnrichment,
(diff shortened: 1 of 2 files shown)


thomaspoignant (approved):
Hey, good catch 🙌 
Thanks a lot for this contribution, I will be available in the next version _(probably this week)_.

PR #1692, opened 2024-03-26
Fix: check for s.AwsConfig when initializing s3Retrieverv2

### LINKED ISSUE #1691: (bug) AwsConfig ignored as of s3Retrieverv2 refactoring
### Observed behaviour

As of 1.24, passing in `AwsConfig` to s3Retrieverv2 won't actually get used. It gets overwritten at Init since the internal `downloader` attribute is nil. This results in our services not being able to initialize the feature flag client since we're setting up our own AwsConfig that's different from what `config.LoadDefaultConfig` will use. 

I believe the issue is the change of checking if `AwsConfig` [is nil previously](https://github.com/thomaspoignant/go-feature-flag/pull/1606/files#diff-72ad8ecde378ae7ac3bfb1dc7a83b19fdd0efaf5b7612b420dbc5c403611405fL66) to checking if [the downloader is nil now](https://github.com/thomaspoignant/go-feature-flag/pull/1606/files#diff-72ad8ecde378ae7ac3bfb1dc7a83b19fdd0efaf5b7612b420dbc5c403611405fR34). I'm not exactly certain though on the intention of the refactor though, so this might not be correct. 

### Expected Behavior

I'd expect passing in AwsConfig to result in that being used in s3retrieverv2.Retriever. Right now I don't see a path that this Config ever gets used. 

### Steps to reproduce

I don't have an independent reproduction at this time, but moving from 1.23 to 1.24 caused our calls to `ffclient.Init` to start failing with an error for: 

```impossible to retrieve the flags, please check your configuration: unable to download item from S3 "feature_flags.yaml", operation error S3: GetObject, resolve auth scheme: resolve endpoint: endpoint rule error, Invalid region: region was not a valid DNS name.```

This is because we set the region directly by calling `config.WithRegion("us-west-2")`, but that gets ignored since the retriever is only using `config.LoadDefaultConfig` and falling back to the ENV variable (I presume) which we don't set.

david-alza:
# Description
Fixes issue with s3Retrieverv2 initialization by checking again for presence of AwsConfig and using that instead of default AWS Config. 

# Changes include

# Closes issue(s)

Resolve #1691 

# Checklist
- [x] I have tested this code. I tested this fork in our own code and it fixed the issue we were seeing before. 
- [ ] I have added unit test to cover this code: I don't believe this is really testable given the current structure of Init. 

### DIFF
--- a/retriever/s3retrieverv2/retriever.go
+++ b/retriever/s3retrieverv2/retriever.go
@@ -33,12 +33,14 @@ type Retriever struct {
 func (s *Retriever) Init(ctx context.Context, _ *log.Logger) error {
 	s.status = retriever.RetrieverNotReady
 	if s.downloader == nil {
-		cfg, err := config.LoadDefaultConfig(ctx)
-		if err != nil {
-			s.status = retriever.RetrieverError
-			return fmt.Errorf("impossible to init S3 retriever v2: %v", err)
+		if s.AwsConfig == nil {
+			cfg, err := config.LoadDefaultConfig(ctx)
+			if err != nil {
+				s.status = retriever.RetrieverError
+				return fmt.Errorf("impossible to init S3 retriever v2: %v", err)
+			}
+			s.AwsConfig = &cfg
 		}
-		s.AwsConfig = &cfg
 		client := s3.NewFromConfig(*s.AwsConfig)
 		s.downloader = manager.NewDownloader(client)
 	}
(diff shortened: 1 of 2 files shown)


thomaspoignant (approved):

thomaspoignant:
Looks good to me, I will test it tomorrow and publish a new version with the patch.

david-alza:
> Looks good to me, I will test it tomorrow and publish a new version with the patch.

Thanks @thomaspoignant. I just added some simple unit tests to cover this. Lemme know if you want me to add anything further.

thomaspoignant (approved):

thomaspoignant:
Thanks a lot for this fix.
I will create a version `v1.24.2` containing the fix.

Issue #1841, opened 2024-05-03
(bug) [relay-proxy] Override configuration via env variables in `retrievers` does not work


thomaspoignant:
### Observed behaviour

When running the relayproxy it is possible to override the configuration file using environment variables.
If the configuration to override is inside an array it is not possible to override it.

This is particularly true when dealing with secrets as mentioned in #1822.

With the current implementation of `koanf` this is not working well, we should consider following what is proposed in this issue https://github.com/knadh/koanf/issues/74.

### Expected Behavior

We want to be able to override a property from the key retrievers with an environment variable.

Something like 
```shell 
export RETRIEVERS_0_HEADERS_AUTHORIZATION=Bearer xxx
```

luanxuechao:
Hi @thomaspoignant , I try to fix it, but I've found a bug in local, when I set yaml in the local:
```
listen: 1031
pollingInterval: 1000
startWithRetrieverError: false
retriever:
  kind: http
  url: https://raw.githubusercontent.com/thomaspoignant/go-feature-flag/main/examples/retriever_file/flags.goff.yaml
retrievers:
   - kind: file
     path: examples/retriever_file/flags.goff.yaml
  - kind: file
    path: examples/retriever_file/flags.goff.yaml

exporter:
  kind: log
enableSwagger: true
authorizedKeys:
  evaluation:
    - apikey1 # owner: userID1
    - apikey2 # owner: userID2
  admin:
    - apikey3
```
the `retrievers` of config can't load from the yaml file.

thomaspoignant:
@luanxuechao from your example I see an indentation issue in your file on the providers.
Can you try this one instead?

```yaml
listen: 1031
pollingInterval: 1000
startWithRetrieverError: false
retriever:
  kind: http
  url: https://raw.githubusercontent.com/thomaspoignant/go-feature-flag/main/examples/retriever_file/flags.goff.yaml
retrievers:
  - kind: file
    path: examples/retriever_file/flags.goff.yaml
  - kind: file
    path: examples/retriever_file/flags.goff.yaml

exporter:
  kind: log
enableSwagger: true
authorizedKeys:
  evaluation:
    - apikey1 # owner: userID1
    - apikey2 # owner: userID2
  admin:
    - apikey3
  ```

Can you double check if this is

thomaspoignant:
I've tried the exact same file as you and I don't reproduce.
What I have done:
[...]

PR #1902, opened 2024-05-21
fix(relay-proxy): Allow overide of array configuration by environment variables

luanxuechao:
# Description

# Closes issue(s)

Resolve [1841](https://github.com/thomaspoignant/go-feature-flag/issues/1841)

# Checklist
- [x] I have tested this code

### DIFF
--- a/cmd/relayproxy/config/config.go
+++ b/cmd/relayproxy/config/config.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
@@ -91,10 +92,14 @@ func New(flagSet *pflag.FlagSet, log *zap.Logger, version string) (*Config, erro
 			log.Error("error loading file", zap.Error(errBindFile))
 		}
 	}
-
+	configMap := k.Raw()
 	// Map environment variables
-	_ = k.Load(env.Provider("", ".", func(s string) string {
-		return strings.ReplaceAll(strings.ToLower(s), "_", ".")
+	_ = k.Load(env.ProviderWithValue("", ".", func(s string, v string) (string, interface{}) {
+		if strings.HasPrefix(s, "RETRIEVERS") || strings.HasPrefix(s, "NOTIFIERS") {
+			_ = loadArrayEnv(s, v, configMap)
+			return s, v
+		}
+		return strings.ReplaceAll(strings.ToLower(s), "_", "."), v
 	}), nil)
 
 	_ = k.Set("version", version)
@@ -321,4 +326,46 @@ func locateConfigFile(inputFilePath string) (string, error) {
 	}
 	return "", fmt.Errorf(
 		"impossible to find config file in the default locations [%s]", strings.Join(defaultLocations, ","))
+}
+
+// Load the ENV Like:RETRIEVERS_0_HEADERS_AUTHORIZATION
+func loadArrayEnv(s string, v string, configMap map[string]interface{}) error {
+	paths := strings.Split(s, "_")
+	for i, str := range paths {
+		paths[i] = strings.ToLower(str)
+	}
+	prefixKey := paths[0]
+	if configArray, ok := configMap[prefixKey].([]interface{}); ok {
+		index, err := strconv.Atoi(paths[1])
+		if err != nil {
+			return err
+		}
+		configItem := configArray[index].(map[string]interface{})
+		keys := paths[2:]
+		currentMap := configItem
+		for i, key := range keys {
+			hasKey := false
+			lowerKey := key
+			for y := range currentMap {
+				if y != lowerKey {
+					continue
+				}
+				if nextMap, ok := currentMap[y].(map[string]interface{}); ok {
+					currentMap = nextMap
+					hasKey = true
+					break
+				}
+			}
+			if !hasKey && i != len(keys)-1 {
+				newMap := make(map[string]interface{})
+				currentMap[lowerKey] = newMap
+				currentMap = newMap
+			}
+		}
+		lastKey := keys[len(keys)-1]
+		currentMap[lastKey] = v
+		configArray[index] = configItem
+		_ = k.Set(prefixKey, configArray)
+	}
+	return nil
 }
--- a/cmd/relayproxy/config/config.go
+++ b/cmd/relayproxy/config/config.go
@@ -97,7 +97,10 @@ func New(flagSet *pflag.FlagSet, log *zap.Logger, version string) (*Config, erro
 	_ = k.Load(env.ProviderWithValue("", ".", func(s string, v string) (string, interface{}) {
 		if strings.HasPrefix(s, "RETRIEVERS") || strings.HasPrefix(s, "NOTIFIERS") {
 			err := loadArrayEnv(s, v, configMap)
-			log.Error("config: error loading array env", zap.String("key", s), zap.String("value", v), zap.Error(err))
+			if err != nil {
+				log.Error("config: error loading array env", zap.String("key", s), zap.String("value", v), zap.Error(err))
+				return s, v
+			}
 			return s, v
 		}
 		return strings.ReplaceAll(strings.ToLower(s), "_", "."), v
--- a/cmd/relayproxy/config/config.go
+++ b/cmd/relayproxy/config/config.go
@@ -2,13 +2,6 @@ package config
 
 import (
 	"fmt"
-	"net/http"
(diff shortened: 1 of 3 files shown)


cmd/relayproxy/config/config.go
```diff
@@ -322,3 +327,45 @@ func locateConfigFile(inputFilePath string) (string, error) {
...
+	for i, str := range paths {
+		paths[i] = strings.ToLower(str)
+	}
+	prefixKey := paths[0]
+	if configArray, ok := configMap[prefixKey].([]interface{}); ok {
+		index, err := strconv.Atoi(paths[1])
+		if err != nil {
+			return err
+		}
+		configItem := configArray[index].(map[string]interface{})
```
thomaspoignant, line 343: Here we can have an error if we try to add an item to the array.

Let's say I want to add a new retriever and not just override it, I will have some env variable like this:
```
RETRIEVERS_1_KIND=github
RETRIEVERS_1_REPOSITORYSLUG=thomaspoignant/go-feature-flag
RETRIEVERS_1_BRANCH=main
RETRIEVERS_1_FILEPATH=testdata/flag-config.yaml
```

And when doing this, the `configArray[index]` will error because the array size is 0 in our case.

cmd/relayproxy/config/config.go
```diff
@@ -91,10 +92,14 @@ func New(flagSet *pflag.FlagSet, log *zap.Logger, version string) (*Config, erro
...
 		}
 	}
-
+	configMap := k.Raw()
 	// Map environment variables
-	_ = k.Load(env.Provider("", ".", func(s string) string {
-		return strings.ReplaceAll(strings.ToLower(s), "_", ".")
+	_ = k.Load(env.ProviderWithValue("", ".", func(s string, v string) (string, interface{}) {
+		if strings.HasPrefix(s, "RETRIEVERS") || strings.HasPrefix(s, "NOTIFIERS") {
+			_ = loadArrayEnv(s, v, configMap)
```
thomaspoignant, line 99: Here we are ignoring the error.
We should at least log the error to be sure that the user has feedback if it does not work as expected.

```suggestion
			err := loadArrayEnv(s, v, configMap)
			log.Error("config: error loading array env", zap.String("key", s), zap.String("value", v), zap.Error(err))
```

thomaspoignant (changes requested):
Hey @luanxuechao thanks a lot for this pull request and this looks super promising.

I've made some small comments, but the main thing I see is that we cannot add a new entry to the array. 

_PS: I have changed the title of the PR to make it more explicit._

cmd/relayproxy/config/config.go
```diff
@@ -91,10 +92,15 @@ func New(flagSet *pflag.FlagSet, log *zap.Logger, version string) (*Config, erro
...
 	}
-
+	configMap := k.Raw()
 	// Map environment variables
-	_ = k.Load(env.Provider("", ".", func(s string) string {
-		return strings.ReplaceAll(strings.ToLower(s), "_", ".")
+	_ = k.Load(env.ProviderWithValue("", ".", func(s string, v string) (string, interface{}) {
+		if strings.HasPrefix(s, "RETRIEVERS") || strings.HasPrefix(s, "NOTIFIERS") {
+			err := loadArrayEnv(s, v, configMap)
+			log.Error("config: error loading array env", zap.String("key", s), zap.String("value", v), zap.Error(err))
```
thomaspoignant, line 100: We should not log the error if we don't have an error.

```suggestion
			if err != nil {
				log.Error("config: error loading array env", zap.String("key", s), zap.String("value", v), zap.Error(err))
			}
```

cmd/relayproxy/config/config.go
```diff
@@ -91,10 +92,15 @@ func New(flagSet *pflag.FlagSet, log *zap.Logger, version string) (*Config, erro
 			log.Error("error loading file", zap.Error(errBindFile))
 		}
 	}
-
+	configMap := k.Raw()
```
thomaspoignant, line 95: If we keep this outside of the function we are not able to pass multiple environment variables.

If we have something like
```
RETRIEVERS_1_KIND=github
RETRIEVERS_1_REPOSITORYSLUG=thomaspoignant/go-feature-flag
RETRIEVERS_1_BRANCH=main
RETRIEVERS_1_FILEPATH=testdata/flag-config.yaml
```

Only the last environment variable will be taken into consideration.

_(check proposal bellow to make it work)_

```suggestion
```

cmd/relayproxy/config/config.go
```diff
@@ -91,10 +92,15 @@ func New(flagSet *pflag.FlagSet, log *zap.Logger, version string) (*Config, erro
...
 	// Map environment variables
-	_ = k.Load(env.Provider("", ".", func(s string) string {
-		return strings.ReplaceAll(strings.ToLower(s), "_", ".")
+	_ = k.Load(env.ProviderWithValue("", ".", func(s string, v string) (string, interface{}) {
+		if strings.HasPrefix(s, "RETRIEVERS") || strings.HasPrefix(s, "NOTIFIERS") {
```
thomaspoignant, line 98:
```suggestion
		if strings.HasPrefix(s, "RETRIEVERS") || strings.HasPrefix(s, "NOTIFIERS") {
			configMap := k.Raw()
```

We move the `configMap` variable inside the function.

thomaspoignant (changes requested):
This version is now accepting env variables outside of the range BUT it does not take into consideration multiple variables in the same object.

Something like
```
RETRIEVERS_1_KIND=github
RETRIEVERS_1_REPOSITORYSLUG=thomaspoignant/go-feature-flag
RETRIEVERS_1_BRANCH=main
RETRIEVERS_1_FILEPATH=testdata/flag-config.yaml
```

With the actual version of the code, it will only take the last env variable.

When we move the location of the `configMap` initialization it will work.
I'll suggest that you double-check and that it works, and it is worth adding a test to validate it.

cmd/relayproxy/config/config.go
luanxuechao: Hi @thomaspoignant, I don't think so, I have reloaded it every time. I also have ut to cover it
https://github.com/thomaspoignant/go-feature-flag/pull/1902/files#diff-9b365e54858c1fbb3914851507d0d67c4b848f841db79c0f460d18f2df974926R387

cmd/relayproxy/config/config_test.go
```diff
@@ -692,3 +692,100 @@ func TestConfig_APIAdminKeyExists(t *testing.T) {
...
+			},
+			wantErr: assert.NoError,
+		},
+	}
+	for _, tt := range tests {
+		os.Setenv("RETRIEVERS_0_HEADERS_AUTHORIZATION", "test")
+		os.Setenv("RETRIEVERS_X_HEADERS_AUTHORIZATION", "test")
+		os.Setenv("RETRIEVERS_1_HEADERS_AUTHORIZATION", "test1")
+		os.Setenv("RETRIEVERS_0_HEADERS_TOKEN", "token")
+		os.Setenv("RETRIEVERS_2_HEADERS_AUTHORIZATION", "test1")
```
thomaspoignant, line 769:
```suggestion
		os.Setenv("RETRIEVERS_0_HEADERS_AUTHORIZATION", "test")
		os.Setenv("RETRIEVERS_X_HEADERS_AUTHORIZATION", "test")
		os.Setenv("RETRIEVERS_1_HEADERS_AUTHORIZATION", "test1")
		os.Setenv("RETRIEVERS_0_HEADERS_TOKEN", "token")
		os.Setenv("RETRIEVERS_2_HEADERS_AUTHORIZATION", "test1")
		os.Setenv("RETRIEVERS_2_HEADERS_X-GOFF-CUSTOM", "custom")
```

cmd/relayproxy/config/config.go
thomaspoignant: Your test does not cover what I wanted to highlight.
If in you test you have 2 entries for the same number (as proposed [here](https://github.com/thomaspoignant/go-feature-flag/pull/1902/files#r1616765918)), you will see that only the last env variable will be taken in consideration.

For example here, you will have only the `RETRIEVERS_2_HEADERS_X-GOFF-CUSTOM` in your configuration and not `RETRIEVERS_2_HEADERS_AUTHORIZATION`.

If you move the `configMap` variable as suggested it will work.

thomaspoignant (commented):
I think we had a misunderstanding on my previous comment.
I hope what I said is clarifying a bit why I think it is not working as expected.

PR #1946, opened 2024-05-31
feat: allow configuring relay-proxy request log verbosity

fals:
## Description
TL;DR;
We are using the `relay-proxy` for a while in our system and we noticed that every request to relay proxy is recorded as log with level=info. Currently this log became a problem, because we generate 1.8 million request every 5 min to relay-proxy. The amount of logs injected in increase as consequence storage and cost.

This PR implements configuring the verbosity of `relay-proxy`, allowing you to log more if needed, or only errors.

## Checklist
- [x] I have tested this code


fals (commented):
allow select verbosity of relay proxy request logging

cmd/relayproxy/api/middleware/zap.go
```diff
@@ -65,9 +66,9 @@ func ZapLogger(log *zap.Logger, config *config.Config) echo.MiddlewareFunc {
 				log.With(zap.Error(v.Error)).Error("Server error", fields...)
 			case n >= 400:
 				log.With(zap.Error(v.Error)).Warn("Client error", fields...)
-			case n >= 300:
+			case n >= 300 && cfg.VerboseRequestLogging:
 				log.Info("Redirection", fields...)
-			default:
+			case cfg.VerboseRequestLogging:
 				log.Info("Success", fields...)
```
thomaspoignant, line 72: We should probably play with the log levels here rather that having to check `VerboseRequestLogging`.

Something like 
```go
case n >= 400:
  log.With(zap.Error(v.Error)).Warn("Client error", fields...)
case n >= 300:
  log.Info("Redirection", fields...)
default:
  log.debug("Success", fields...)

```

thomaspoignant (commented):
Hello @fals,
thanks a lot for this PR and I think you are right but we should probably be better in log management in general.

Instead of the `VerboseRequestLogging` I would prefer to have a property called `LogLevel` to set the level of the logs we want for the application.
The levels should be the classic ones `debug`, `info`, `warn`, `error`.

By setting this new option, we can also maybe deprecate the [`debug`](https://github.com/thomaspoignant/go-feature-flag/blob/3456858110bd4a5495fb4dfb12b13d98a90d3b37/cmd/relayproxy/config/config.go#L132) that is a bit too trivial here.

cmd/relayproxy/api/middleware/zap.go
fals: @thomaspoignant debug won't help either, I want to be able to fully remove those request logs.

fals: @thomaspoignant I believe I misunderstood the first comment, playing with the log level will help, but only if we can exclude lower levels. Would that work? I can simply flip the flag as you suggested as `LogLevel` and in the secton of code filter the ones that are lower than specified.

thomaspoignant: Yes sorry if it was not clear, you can tell `uber/zap` to log only regarding to the `LogLevel` configured.

fals: @thomaspoignant nice! I'm pushing the change with the LogLevel instead 👍🏻

fals: @thomaspoignant I'm sure that what I suggested in the PR is enough, probably have to update something else. Regarding the Debug flag, idk what is the process you use for deprecating fields, but the log level is enough enable verbosity as you proposed.

cmd/relayproxy/api/middleware/zap.go
```diff
@@ -30,10 +31,10 @@ func DebugSkipper(_ echo.Context) bool {
 }
 
 // ZapLogger is a middleware and zap to provide an "access log" like logging for each request.
-func ZapLogger(log *zap.Logger, config *config.Config) echo.MiddlewareFunc {
+func ZapLogger(log *zap.Logger, cfg *config.Config) echo.MiddlewareFunc {
```
thomaspoignant, line 34: Good catch 👍

cmd/relayproxy/config/config.go
```diff
@@ -32,6 +34,8 @@ var DefaultRetriever = struct {
 	GitBranch:  "main",
 }
 
+const DefaultLogLevel = "debug"
```
thomaspoignant, line 37: Not sure if we should be by default to `debug`, I think `info` would be better.

cmd/relayproxy/log/logger.go
```diff
@@ -1,10 +1,26 @@
...
+
+func InitLogger() *Logger {
+	logger := &Logger{Atom: zap.NewAtomicLevel()}
+
+	encoderCfg := zap.NewProductionEncoderConfig()
+	logger.ZapLogger = zap.New(zapcore.NewCore(
+		zapcore.NewJSONEncoder(encoderCfg),
+		zapcore.Lock(os.Stdout),
+		logger.Atom,
+	))
```
thomaspoignant, line 23: I'll prefer to avoid having to set any configuration manually except the level and rely on the `NewProductionConfig()` helper.
```suggestion
        prodConfig := zap.NewProductionConfig()
        prodConfig.Level = logger.Atom
	logger.ZapLogger = prodConfig.Build()
```

thomaspoignant (changes requested):
Sorry for the time it took for me to review it.
I added some comments about:
- the removal of the `debug` option.
- the way you init `zap`.
- the name of the option that is not using camelcase in the config file.

Despite that it looks super good 👍

fals:
> Sorry for the time it took for me to review it. I added some comments about:
> 
> * the removal of the `debug` option.
> * the way you init `zap`.
> * the name of the option that is not using camelcase in the config file.
> 
> Despite that it looks super good 👍

cmd/relayproxy/main.go
```diff
@@ -48,37 +48,40 @@ func main() {
 	_ = f.Parse(os.Args[1:])
 
 	// Init logger
-	zapLog := log.InitLogger()
-	defer func() { _ = zapLog.Sync() }()
+	logger := log.InitLogger()
```
fals, line 51: @thomaspoignant any tip to satisfy this coverage? The test for main is empty, just a dummy test there, and all new code I introduced is flagged, so I believe I have to find where exactly you test this cases. Could you guide me?

thomaspoignant: Most of the time I don't test the main function.
So I guess this is not really a problem here.

fals: @thomaspoignant let me know them what else needs to change to push this forward. The codecoverage isn't passing, that's why I asked the question.

cmd/relayproxy/config/config.go
```diff
@@ -112,6 +119,14 @@ func New(flagSet *pflag.FlagSet, log *zap.Logger, version string) (*Config, erro
 	if errUnmarshal != nil {
 		return nil, errUnmarshal
 	}
+
+	if proxyConf.Debug && proxyConf.LogLevel == "" {
+		log.Warn(
+			"Option Debug that you are using in your configuration file is deprecated" +
+				"and will be removed in future versions." +
+				"Please use logLevel: debug to continue to run the relay-proxy with debug logs.")
+	}
```
thomaspoignant, line 128: `proxyConf.LogLevel == ""` will never be true because you have a default value.


```suggestion
	if proxyConf.Debug {
		log.Warn(
			"Option Debug that you are using in your configuration file is deprecated " +
				"and will be removed in future versions." +
				"Please use logLevel: debug to continue to run the relay-proxy with debug logs.")
	}
```

cmd/relayproxy/config/config.go
```diff
@@ -387,3 +413,26 @@ func loadArrayEnv(s string, v string, configMap map[string]interface{}) error {
 	}
 	return nil
 }
+
+func (c *Config) IsDebugEnabled() bool {
+	if c == nil {
+		return false
+	}
+	return c.LogLevel == "debug" || c.Debug
```
thomaspoignant, line 421: Checking in lowercase to be ready for any configuration string.
```suggestion
	return strings.ToLower(c.LogLevel) == "debug" || c.Debug
```

thomaspoignant (approved):
I have spotted 2 minor things and I have pushed 1 more commit.
I will merge the PR.

Thanks a lot @fals for your contribution, this will be available in the next version of GO Feature Flag.
To continue improving the logging of GOFF I have also opened this PR #1955 that will support levelled log inside the core go module of GOFF.

Issue #2196, opened 2024-08-05
(feature) [HELM] Allow for creating Role/RoleBinding for loading flags from ConfigMap


yafanasiev:
### Requirements

Hi! First of all, thanks for a great project!

We are setting up go-feature-flag to load flag definitions from Kubernetes ConfigMap (relay-proxy itself is deployed in Kubernetes as well) and to make it work we have to add Role and RoleBinding to ServiceAccount created by chart (which is kind of expected as well), however might be not as convenient since you can't have Helm managing all the required resources.

Would you be open to adding those RBAC resources to the chart? Or at least a generic template allowing to pass any YAML to be rendered into resource. I would be happy to do a PR for this.

Just for reference, this is how sample RBAC setup looks like:

```yaml
kind: Role
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: feature-flags-relay-proxy
rules:
- apiGroups: [""]
  resources:
  - configmaps
  verbs:
  - get
---
kind: RoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: feature-flags-relay-proxy
subjects:
- kind: ServiceAccount
  name: feature-flags-relay-proxy
roleRef:
  kind: Role
  name: feature-flags-relay-proxy
  apiGroup: ""
 ```

thomaspoignant:
Hello @yafanasiev,
Yes, it sounds like a good thing to have, but if you work on it just keep in mind the backwards compatibility and it should continue to work as before for people upgrading the charts.

Adding a generic template for including custom YAML seems to be the most appropriate approach to avoid breaking changes. But I am happy to have your point of view on this.

yafanasiev:
I went the generic manifests path, since indeed my use-case may not be as wide-spread and overloading chart with very specific values will just increase the complexity. Please review when you have time https://github.com/thomaspoignant/go-feature-flag/pull/2198

PR #2198, opened 2024-08-06
feat(relay-proxy-helm): add ability to deploy extra manifests

yafanasiev:
## Description
Adds an ability to deploy arbitrary manifests by passing them through the values, with support for templates.
Additionally, added annotation on deployment with configuration ConfigMap checksum so pods are automatically rotated when config is changed, removing the need for manual restart.

## Closes issue(s)

Resolve https://github.com/thomaspoignant/go-feature-flag/issues/2196

## Checklist
- [x] I have tested this code


thomaspoignant:
Hey, thanks for the PR.
Since I had doubt about this (because it gives the opportunity to anyone to push what they want in the chart), I asked the helm community and they suggested another approach (see [here](https://kubernetes.slack.com/archives/C0NH30761/p1722950773799489)).

What they suggest is:
>I think values.yaml should not have extra manifests, rather if someone has requirements to create extra resources, he or she should use this chart as dependency and create their own chart with a template folder having extra manifests.

What do you think about that approach @yafanasiev?

yafanasiev:
@thomaspoignant I can see how it could lead to some issues if you provide broken/conflicting manifests, but that's completely on the end user of the chart. Essentially, chart itself with the default values produces no extra manifests, and user is making a conscious decision populating those `extraManifests`, and it all happens in the context of his own deployment. Creating a separate chart is quite a step (and you also have to keep in mind added complexity on how helm dependencies work). And for me the culprit here is using the chart with FluxCD/ArgoCD - it is easy to set up a Helm release with a correct set of values, even if some of those are "escape hatches" rather then adding some manifests alongside.

thomaspoignant (approved):
@yafanasiev I get it, and as I mentioned in the issue I don't have a strong opinion on this one.
We can go with it, I will merge it and it will be available in the next version.

PR #2218, opened 2024-08-13
feat: Implement dynamic targeting key

### LINKED ISSUE #2135: (feature) Support for Custom Assignment/Targeting Keys
## Description
_This feature was originally asked by Ricardo Tomasi in this [slack conversation](https://gophers.slack.com/archives/C029TH8KDFG/p1720704534731819)._
 
This feature request proposes to do the percentages rollout based on another field than the targeting key.
This would allow for greater flexibility in defining target audiences for feature flags, specifically facilitating scenarios where flags need to be assigned based on attributes other than the user ID.

## Current Behavior:
Currently, GO Feature Flagassign variations are based on the targetingKey by default. This presents limitations when targeting needs to be based on other attributes, like team ID.

This can cause some limitations, especially when using a client-side SDK that uses a single evaluation context to evaluate all the flags and it is not possible to change it per flags.

> [!NOTE]
> In a percentage split the way we affect users to the variation is something like 
> ```
> (targetingKey + flag name) = 100000
> 
> // depending on the percentage we have buckets like this
> | variation A (0..5000) | variation B (5000...40000) | variation C (40000...100000) |
> ``` 

## Desired Behavior:
**Dynamic Targeting Key**: Introduce a new field within the feature flag configuration to allow choosing which field from the evaluation context will be used to do the variation affectation in a percentage split of the traffic.

# Proposed solution

To support this feature we can add a new field in the rules called `bucketingKey` that will allow to specify which key from the evaluation context will be used to split the traffic.

It will look like this:

```yaml
my-flag:
  variation:
     enabled: true
     disabled: false
  defaultRule:
     percentage:
       enabled: 10
       disabled: 90
     bucketingKey: teamId
```

Now when evaluating a flag with an evaluation context looking like this:
```json
{
  "targetingKey": "04b86a91-101d-4b16-87c7-ecb3af38a869",
  "email": "contact@gofeatureflag.org"
  "teamId": 156
}
```

Instead of affecting like this :
```
(targetingKey + flag name) = 100000
```

We will do 
```
(teamId + flag name) = 100000
```

⚠️ If for any reason the you `bucketingKey` is not available in the evaluation context, we will use the `targetingKey` as a fallback method to affect the right variation.

ricardobeat:
## Description

Introduces the `bucketingKey` field in flag configuration. It allows the targeting key to be dynamically overwritten at evaluation time by another value coming from the context. See related issue for more info.

This is a non-breaking change.

## Closes issue(s)

Resolve #2135

## Checklist
- [x] I have tested this code


ricardobeat:
Since rules don't have access to the parent flag, I had to pass the `key` around, there might be a better way to achieve this?

internal/flag/error_code.go
```diff
@@ -12,6 +12,7 @@ const (
 	ErrorCodeGeneral             ErrorCode = "GENERAL"
 	ErrorCodeInvalidContext      ErrorCode = "INVALID_CONTEXT"
 	ErrorCodeTargetingKeyMissing ErrorCode = "TARGETING_KEY_MISSING"
+	ErrorCodeBucketingKeyMissing ErrorCode = "BUCKETING_KEY_MISSING"
```
thomaspoignant, line 15: I am a bit worried about this one because it is not part of the valid OpenFeature error code _(see list [here](https://openfeature.dev/specification/types/#error-code))._

It means that every Go Feature Flag provider will fall back to `GENERAL` and it will not give any particular information to the user of this flag on why we always use the default value provided.

thomaspoignant (commented):
Hey @ricardobeat, thanks a lot for creating this pull request! It helps me understand a bit better what we were discussing.

If I'm correct, you want the evaluation to fail and return to the default value provided in the evaluation context if the bucketingKey is not found.

This is an interesting idea and a simple way to avoid having to fall back on the `targetingKey`. However, I'm concerned about the new `ErrorCode` we're sending because it's not a known one from OpenFeature (see list [here](https://openfeature.dev/specification/types/#error-code)).

I've also reviewed the different documentation you shared in the [Slack channel](https://gophers.slack.com/archives/C029TH8KDFG/p1720777644977279?thread_ts=1720704534.731819&cid=C029TH8KDFG). It seems there's no consensus on whether to fall back to default behavior or fail.

Optimizely falls back to the targetingKey (see [here](https://docs.developers.optimizely.com/full-stack-experimentation/v2.0/docs/assign-variations-with-bucketing-ids)).
> If you don't pass a bucketing ID to the attributes parameter, users are bucketed by user IDs, which is the default method.

I haven't done a deep review of the code yet but it looks good, I am just not sure if falling is a good idea when the `bucketingKey` required by the flag is not set in the evaluation context 🤔.

ricardobeat:
> Optimizely falls back to the targetingKey
>  not sure if falling is a good idea when the bucketingKey required by the flag is not set in the evaluation context 

I couldn't find Optimizely's reasoning for this. Here is my attempt at illustrating the problem:

image

The user in _org1_ receives the `on` variant; let's imagine this is a new "Team updates" section in their app. The other user in the organization is in the `off` variant and will never see the feature.

Not only this defeats the purpose of the feature test, but when we set up analytics for the A/B test, we are going to measure things like interaction numbers also bucketed by org. Having users randomly assigned to different variants within the same org means the data won't be useful, or worse, will be silently biased if this (missing key) happens only for a small % of users.

Personally I don't see a reason to do it any other way, but would be happy if this is at least configurable behaviour.

internal/flag/internal_flag.go
```diff
@@ -68,6 +72,17 @@ func (f *InternalFlag) Value(
 		maps.Copy(evaluationCtx.GetCustom(), flagContext.EvaluationContextEnrichment)
 	}
 
+	key, keyError := f.GetTargetingKey(evaluationCtx)
+
+	if keyError != nil {
+		return flagContext.DefaultSdkValue, ResolutionDetails{
+			Variant:   VariationSDKDefault,
+			Reason:    ReasonError,
+			ErrorCode: ErrorCodeBucketingKeyMissing,
```
thomaspoignant, line 81: One thing we can do to be explicit about the error is to add an `ErrorDetails` field in the `ResolutionDetails` object that explains the error.

This is also following what OpenFeature is doing with the error message https://openfeature.dev/docs/reference/concepts/evaluation-api/#evaluation-details-structure-fields

internal/flag/internal_flag.go
```diff
@@ -340,6 +355,20 @@ func (f *InternalFlag) GetVariationValue(name string) interface{} {
 	return nil
 }
 
+func (f *InternalFlag) GetTargetingKey(ctx ffcontext.Context) (string, error) {
```
thomaspoignant, line 358: I will rename this function `GetBucketingKey` because this is why we are using it, we want the field used for the bucketing.
```suggestion
func (f *InternalFlag) GetBucketingKey(ctx ffcontext.Context) (string, error) {
```

internal/utils/serializer.go
```diff
@@ -14,6 +14,6 @@ func ContextToMap(ctx ffcontext.Context) map[string]interface{} {
 		userCopy[key] = value
 	}
 	userCopy["anonymous"] = ctx.IsAnonymous()
-	userCopy["key"] = ctx.GetKey()
+	userCopy["key"] = key
```
thomaspoignant, line 17: As mentioned in the issue, I'd prefer not to do this.
`key` and `targetingKey` are special fields in the evaluation context and we should keep them like that.

For this feature, the important part is to be sure that we split the users based on another field it should not have an impact on the way to write rules.

This part will change how you write your rules and may lead to misleading configuration of your flags.

Let's say you have this flag
```yaml
my-flag:
  variation:
     enabled: true
     disabled: false
  defaultRule:
     percentage:
       enabled: 10
       disabled: 90
  bucketingKey: teamId
```
and this evaluation context:
```json
{
  "targetingKey": "f1f0d804-f23f-4c93-a935-6e04ce2cf696",
  "teamId": 567,
  "email": "contact@gofeatureflag.org"
}
```

If you have to add a rule inside the flag based on the `teamId` you can go with
```yaml
# ...
  targeting:
     - query: teamId eq "567"
       variation: enabled
# ...
```
I don't see the adding value to also put `teamId` value as `key`.

thomaspoignant (commented):
You are right, falling back to the default value is probably the best option here.
I've put some comments in the PR, especially about replacing the `key` in the context where I don't see the adding value.

If we want to move forward with this PR, I think we should address the documentation part because it will be necessarily to explain well how it works.
_If you are lost with the doc, I can always help you._

internal/flag/error_code.go
ricardobeat: > missing something from the context that is not a targeting key

It is a targeting key though, in this case. I'm tending towards `ErrorCodeTargetingKeyMissing` - looking at the flag definition and noticing it has a `bucketingKey` makes it obvious what happened - if the error says 'invalid context' I think it's a much wider surface of things that could have gone wrong

ricardobeat:
@thomaspoignant I've done a bit of refactoring according to the comments above:

- renamed `GetBucketingKey`
- removed the non-standard error
- added `ErrorDetails` (note: OpenFeature refers to it as 'error message' but OFREP has `ErrorDetails` instead)

I also took a first pass at documentation. Let me know what you think.

thomaspoignant (changes requested):
Hey @ricardobeat sorry I have been on vacation for some weeks this is why I was a bit quiet.

I think we are close to the solution with this PR, my only remarks is about the tests that are not testing what the change of key is affecting _(in percentage-based evaluation)_.

As soon as we have a good test for that I think we will be ready to merge.

thomaspoignant:
@ricardobeat do you need help finishing this PR?

internal/flag/internal_flag_test.go
```diff
@@ -1710,6 +1711,111 @@ func TestInternalFlag_Value(t *testing.T) {
...
+			want1: flag.ResolutionDetails{
+				Variant:   "variation_A",
+				Reason:    flag.ReasonDefault,
+				Cacheable: true,
+				Metadata: map[string]interface{}{
+					"region": "nl-nh",
+				},
+			},
+		},
+	}
```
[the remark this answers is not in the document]
ricardobeat: > A good test here could be to have a percentage-based evaluation, set a targetingKey that affect to 1 variant and a teamId that affect to another variant.

I'm not sure what to do here

When I was overwriting the `key`, this was actually verifying that _any_ rules would be applied based on that. Now, if a BucketingKey is set, since the `key` remains the same, as far as I can tell there is no way to test that it is actually using the custom bucketing key.  Only that rules (like `teamId eq xxx`) see the attribute, but that's kind of orthogonal to the change.

ricardobeat:
@thomaspoignant sorry, I was also on holiday the week before and just got around to reviewing this. Need some help with figuring out the tests indeed, commented above 🙏

thomaspoignant (approved):

PR #2257, opened 2024-08-23
feat: Add initial delay for polling feature configuration

samuolis:
## Description

- Added `pollingDelayInMillis` in initialize options, to declare delay for the first request of feature flags
- Added tests to cover this case

This should solve the issue of getting feature flags before the first polling even happened (Default delay is 5 mins, it is okay for polling but not okay for waiting for initial data)

## Checklist
- [x] I have tested this code


samuolis:
@thomaspoignant Do this library need code duplication check on tests, because it is falling on already existing tests

thomaspoignant:
@samuolis thanks for your pull request, it fixes an important problem from the provider.

I am currently away (taking some vacation), but I will look closely to your PR early September and we will be able to release a new version of the provider with this addition.

For the code duplication, this should not be a problem for test, I will look at the CI to avoud failling on such cases.

thomaspoignant (commented):
Hi @samuolis,

 thanks for your contribution! I appreciate you taking the time to submit a pull request.

I'd like to understand a bit more about the context of your changes in this PR. While reviewing the code, I noticed that the initial API call to retrieve flags happens before polling starts.
https://github.com/thomaspoignant/go-feature-flag/blob/46aeea6aca273dcb0e9b24f3ac1074cdf6a874a0/openfeature/providers/kotlin-provider/gofeatureflag-kotlin-provider/src/main/java/org/gofeatureflag/openfeature/ofrep/OfrepProvider.kt#L58
This design approach ensures that flags are available immediately, even before the first polling cycle.

Could you elaborate on the specific issue you're trying to address with this pull request? Understanding your thought process will help me provide more relevant feedback.

If you have any questions or need further clarification on the current implementation, feel free to ask! I'm happy to discuss this further.

samuolis:
Sorry for such a delay, I missed this line somehow. I now see what was the problem, somehow the first API call was getting cancelled, now it works perfectly

PR #2292, opened 2024-08-29
feat(relay-proxy-helm): add ability render ingress host as a template

yafanasiev:
## Description

Allows using Helm template variable inside Ingress hostname. We use relay proxy as a part of our ephemeral development environments, so ability to dynamically render the hostname based on namespace means we don't have to change it for every new environment.

## Checklist
- [x] I have tested this code


thomaspoignant (approved):

thomaspoignant (approved):
@yafanasiev Thanks for this addition.
We will release it in the next version of GO Feature Flag.

PR #2365, opened 2024-09-16
fix(python-provider): hook error

### LINKED ISSUE #2366: (bug) <`error` method of `DataCollectorHook` class in python-provider use a non declared variable `details`>
### Observed behaviour

I used the sample program for `python-provider` without starting the relay proxy. I noticed from the log, an error is thrown in `error` method, saying that the `details` variable does not exist. From the code, I can see that it is a copy and paste of the `after` method which has `details` passed as an argument. The `error` method does not have this parameter as an input, so the error was thrown.
This is a syntactical error.

### Expected Behavior

The problem should not throw a syntatical error.

### Steps to reproduce

Run the python sample at [here](https://github.com/thomaspoignant/go-feature-flag.git) from the go-feature-flag doc without started a relay proxy.

The error should be able to be observed from the log.

tungntpham:
Remove the use of `details`. It looks like this has been a copy and paste error.

In the case of evaluation error, regardless of whether the flag has been cached, we should still want to collect an error event

## Description
- What was the problem?
  - The `error` method of `DataCollectorHook` use a non declared variable `details`. This seems to be a copy and paste of the `after` method.
- How it is resolved?
  - Remove the use of `details` in the method
- How can we test the change?
  - Added an unit test
- If there are breaking changes, please describe them in detail and why we cannot avoid them.
  - Not a breaking change

## Closes issue(s)
  
Resolve #2366 

## Checklist
- [x] I have tested this code


thomaspoignant (approved):
Good catch, this is indeed a copy/paste error.
I will release a fix version for the python provider and it will be part of it.

thomaspoignant:
A version `0.2.1` has been released.

thomaspoignant:
@tungntpham A version `0.2.1` has been released.

PR #2377, opened 2024-09-20
fix: add BucketingKey to DTO converter functions

ricardobeat:
## Description

Was pulling my hair trying to figure out why the new bucketing didn't work on the server. Turns out we forgot to add it to the converters, so it gets lost in the cache initialization phase.

Tests don't cover this, should they?

## Checklist
- [x] I have tested this code


ricardobeat:
On a separate note, I noticed that every time I save a file, the imports get re-organized. Is `gofmt` not in use for the project?

thomaspoignant:
There is clearly a missing test, I will add it before merging this PR and releasing a fix version.

PR #2378, opened 2024-09-20
chore: reorder imports in a consistent way

thomaspoignant:
## Description
The project had no safe guards in how the order of imports.
In this PR:
- Adds `gofmt` and `gci` in the linter.
- Reorder all the imports in a consistent way.
- Add auto-fix for `pre-commit` to be sure that we have the right order all the time.

## Checklist
- [x] I have tested this code

### DIFF
--- a/cmd/relayproxy/config/config_test.go
+++ b/cmd/relayproxy/config/config_test.go
@@ -2,13 +2,13 @@ package config_test
 
 import (
 	"fmt"
-	"github.com/stretchr/testify/require"
 	"io"
 	"os"
 	"testing"
 
 	"github.com/spf13/pflag"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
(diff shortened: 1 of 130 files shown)


cmd/relayproxy/config/config_test.go
```diff
@@ -750,35 +750,35 @@ func TestMergeConfig_FromOSEnv(t *testing.T) {
...
-					config.RetrieverConf{
+					{
 						HTTPHeaders: map[string][]string{
 
-							"authorization": []string{
+							"authorization": {
 								"test1",
 							},
-							"x-goff-custom": []string{
+							"x-goff-custom": {
```
thomaspoignant, line 781: this change is because of `gofmt`

thomaspoignant, line 781: this change is because of `gofmt`

Issue #2396, opened 2024-09-20
(feature) Request body logging


augustinasrce:
### Requirements

In order to properly debug request body/evaluation context is needed.

PR #2397, opened 2024-09-20
feat: Request body logging in debug level

augustinasrce:
## Description

Please add a description of what your pull request is doing.
 - What was the problem?
>  In order to debug, request body is very helpful.
 - How it is resolved?
>  Introduced request body logging
 - How can we test the change?
> Enabled/disable debug log level and check logs. Unit tests also available.
 - If there are breaking changes, please describe them in detail and why we cannot avoid them.
> None

## Closes issue(s)
 
 #2396

## Checklist
- [x] I have tested this code


cmd/relayproxy/api/server.go
```diff
@@ -53,6 +53,7 @@ func (s *Server) initRoutes() {
 	s.apiEcho.HidePort = true
 	s.apiEcho.Debug = s.config.IsDebugEnabled()
 	s.apiEcho.Use(custommiddleware.ZapLogger(s.zapLog, s.config))
+	s.apiEcho.Use(custommiddleware.BodyLogger(s.zapLog, s.config))
```
thomaspoignant, line 56:
```suggestion
	s.apiEcho.Use(middleware.BodyDumpWithConfig(middleware.BodyDumpConfig{
		Skipper: func(c echo.Context) bool {
			return !s.zapLog.Core().Enabled(zap.DebugLevel)
		},
		Handler: func(_ echo.Context, reqBody []byte, _ []byte) {
			log.Debug("Request info", zap.ByteString("request_body", reqBody))
		},
	}))
```

cmd/relayproxy/api/middleware/body.go
```diff
@@ -0,0 +1,19 @@
...
+	"go.uber.org/zap"
+)
+
+func BodyLogger(log *zap.Logger, cfg *config.Config) echo.MiddlewareFunc {
+	skipper := func(_ echo.Context) bool { return !cfg.IsDebugEnabled() }
+
+	return middleware.BodyDumpWithConfig(middleware.BodyDumpConfig{
+		Skipper: skipper,
+		Handler: func(_ echo.Context, req []byte, _ []byte) {
+			log.Debug("Request", zap.ByteString("body", req))
```
thomaspoignant, line 16: I am not a fan of using `body` as the property name here I would go more with `request_body`, so we can evolve this log in the future.

```suggestion
			log.Debug("Request info", zap.ByteString("request_body", reqBody))
```

cmd/relayproxy/api/middleware/body.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/thomaspoignant/go-feature-flag/cmd/relayproxy/config"
+	"go.uber.org/zap"
+)
+
+func BodyLogger(log *zap.Logger, cfg *config.Config) echo.MiddlewareFunc {
+	skipper := func(_ echo.Context) bool { return !cfg.IsDebugEnabled() }
+
+	return middleware.BodyDumpWithConfig(middleware.BodyDumpConfig{
+		Skipper: skipper,
+		Handler: func(_ echo.Context, req []byte, _ []byte) {
```
thomaspoignant, line 15:
```suggestion
		Handler: func(_ echo.Context, reqBody []byte, _ []byte) {
```
just to be sure this is not misleading and people knows that this does not contains the full request.

thomaspoignant (commented):
Hey @augustinasrce thanks for your pull request.

This is a great addition to log the request body, but since we are using an existing middleware I am not sure why we need our own file in the middleware section, I would prefer that we include directly the middleware in the `server.go` file as suggested.

augustinasrce:
> Hey @augustinasrce thanks for your pull request.
> 
> This is a great addition to log the request body, but since we are using an existing middleware I am not sure why we need our own file in the middleware section, I would prefer that we include directly the middleware in the `server.go` file as suggested.

Idea was to keep things simple, without big nested structures (in case it gets bigger).  Since it's quite small at the moment I'll update PR as You suggested.

thomaspoignant (approved):

PR #2422, opened 2024-09-24
feat: add deprecation log for v0 flags

### LINKED ISSUE #2399: Log error when v0.x.x flag format is used
> [!WARNING]  
> As mention in the issue #2398 we have decided to drop the support of the `v0.x.x` flag format. 

Since we want to ease this transition for people still using the original format, we should start by introducing an **ERROR** log explaining that new versions of GO Feature Flag will no longer support this format.

The log should be:

```
ERROR: your flag <flag name> is using GO Feature Flag flag format v0.x.x. The support of this format will be removed soon, please consider migrating to the v1 format (see https://gofeatureflag.org/docs/tooling/migrate_v0_v1).
```

thomaspoignant:
## Description
add deprecation logger for v0 flags

## Closes issue(s)

Resolve #2399

## Checklist
- [x] I have tested this code

### DIFF
--- a/internal/cache/in_memory_cache.go
+++ b/internal/cache/in_memory_cache.go
@@ -64,11 +64,11 @@ func (fc *InMemoryCache) All() map[string]flag.Flag {
 }
 
 func (fc *InMemoryCache) Init(flags map[string]dto.DTO) {
-	cache := make(map[string]flag.InternalFlag, 0)
+	cache := make(map[string]flag.InternalFlag)
 	for key, flagDto := range flags {
-		flagToAdd := flagDto.Convert()
+		flagToAdd := flagDto.Convert(fc.Logger, key)
 		if err := flagToAdd.IsValid(); err == nil {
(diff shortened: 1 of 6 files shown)


internal/cache/in_memory_cache.go
```diff
@@ -64,11 +64,11 @@ func (fc *InMemoryCache) All() map[string]flag.Flag {
...
 
 func (fc *InMemoryCache) Init(flags map[string]dto.DTO) {
-	cache := make(map[string]flag.InternalFlag, 0)
+	cache := make(map[string]flag.InternalFlag)
 	for key, flagDto := range flags {
-		flagToAdd := flagDto.Convert()
+		flagToAdd := flagDto.Convert(fc.Logger, key)
 		if err := flagToAdd.IsValid(); err == nil {
-			cache[key] = flagDto.Convert()
+			cache[key] = flagToAdd
```
thomaspoignant, line 71: This is fixing a double convertion issue.

Issue #2452, opened 2024-10-03
(doc) Improve progressive rollout documentation.


thomaspoignant:
### Change in the documentation

The current documentation for progressive rollout is misleading and hard to understand.
https://gofeatureflag.org/docs/configure_flag/rollout/progressive

We should rework it to explain how progressive rollout works.
This [slack thread](https://gophers.slack.com/archives/C029TH8KDFG/p1727942268907469) explains some of the errors of understanding that could happen with the actual documentation.

PR #2456, opened 2024-10-06
docs: update progressive rollout docs

Light2Dark:
## Description

Update the example used and add a description of how the configuration works. (updated)
Screenshot 2024-10-06 at 6 46 12 PM

## Closes issue(s)

Resolve https://github.com/thomaspoignant/go-feature-flag/issues/2452

## Checklist
- [ ] I have tested this code
- [X] I have updated the documentation (`README.md` and `/website/docs`)

### DIFF
--- a/website/docs/configure_flag/rollout/progressive.mdx
+++ b/website/docs/configure_flag/rollout/progressive.mdx
@@ -5,9 +5,6 @@ import TabItem from '@theme/TabItem';
 
 A **progressive rollout** allows you to increase the percentage of your flag over time.
 
-You can select a **release ramp** where the percentage of your flag will increase progressively between the start date
-and the end date.
-
 ## Example
 
 <Tabs groupId="code">
@@ -22,13 +19,13 @@ progressive-flag:
 # highlight-start
     progressiveRollout:
       initial:
-        variation: variationB
+        variation: variationA
         percentage: 0
-        date: 2021-03-20T00:00:00.1-05:00
+        date: 2024-01-01T05:00:00.100Z
       end:
         variation: variationB
         percentage: 100
-        date: 2021-03-21T00:00:00.1-05:00
+        date: 2024-01-05T05:00:00.100Z
 # highlight-end
 ```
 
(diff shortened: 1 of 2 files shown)


Light2Dark:
Hey @thomaspoignant , I faced this issue and wanted to contribute to the docs. I think it might be worth to explain a little more on changing the percentage field. However, it is unintuitive. The following is my proposal to be added to the **Example** section:

### Using the percentage field (advanced)
If you do not intend to completely rollout a feature, you can modify the percentage field.

```yaml
    progressiveRollout:
      initial:
        variation: variationA
        percentage: 20   
        date: 2024-01-01T05:00:00.100Z
      end:
        variation: variationB
        percentage: 80
        date: 2024-01-05T05:00:00.100Z
```

- Before the `initial` date, the flag will return **variationB 20% of the time** and **variationA 80% of the time**.
- Between the `initial` and `end` date, the flag will progressively change from returning variationA to variationB.
- After the `end` date, the flag will return **variationB 80% of the time** and **variationA 20% of the time**.

website/docs/configure_flag/rollout/progressive.mdx
```diff
@@ -71,27 +68,30 @@ variationA = "A"
...
+|------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| **`percentage`**<br/>*(optional)*  | It represents the ramp of progress, at which level the flag starts (`initial`) and ends (`end`).<br/>**Default: `initial` = `0` and `end` = `100`**.                                                                                                                                                                                                                                                                                                                                                                                              |
```
thomaspoignant, line 97: We probably need to change this table with explaining what should be in initial and end.

thomaspoignant (commented):
@Light2Dark thanks a lot for this improvement of the documentation this is really great.

I agree that your proposal for the example is better to understand how it works, just to be sure you want to have 2 different examples? One without percentage _`(which means that we go from 0 to 100)_ and one with it?
I like your proposal, the bullet list helps a lot to understand better how it works.

The only thing I would change is the `Configuration fields` table that is not reflecting the format, I would prefer to have an entry for `initial` and one for `end`.

Light2Dark:
> I agree that your proposal for the example is better to understand how it works, just to be sure you want to have 2 different examples? One without percentage _`(which means that we go from 0 to 100)_ and one with it? I like your proposal, the bullet list helps a lot to understand better how it works.

Yeap, 2 examples. One is a simple use case, and another for more advanced use. Let me make some changes and we can review then :)

> The only thing I would change is the `Configuration fields` table that is not reflecting the format, I would prefer to have an entry for `initial` and one for `end`.

Noted, let me add to this.

thomaspoignant (approved):

PR #2480, opened 2024-10-08
feat: update S3 client initialization for retriever and exporter

### LINKED ISSUE #2451: (change) Change the way the S3 client is instantiated in the exporter/retriever
### Motivation

Currently, `AwsConfig` is the only parameter used to configure the S3 client to use (while AWS offers much more)
https://github.com/thomaspoignant/go-feature-flag/blob/main/exporter/s3exporterv2/exporter.go#L71
This limits the functionality that AWS provides.

In my case, I need to set some additional configurations for the S3 client such as `EndpointResolverV2` as well as `ForcePathStyle` to facilitate inter-container communication when using localstack in local development.
And I can't find a way to set these in `AwsConfig`.

More specifically...
Suppose I set the alias for my localstack in the network as `localhost.localstack.cloud`
Without configuring the above settings, the s3 client will default to the following path to put the object
`{bucket-name}.s3.localhost.localstack.cloud:4566`
It will always resolve to `127.0.0.1`, which is wrong and the expected value should be the IP of the localstack on the network.
This will be solved by forcing the S3 client to work with path style instead of domain
So, everything would be resolved if I could pass additional configuration when initializing the S3 client

This is what I encountered, and there are likely to be many other requirements related to configuring how the S3 client works
So it makes sense to provide as many of them as possible.

### Requirements

Add configuration parameters to fully support the S3 configuration provided by AWS
https://github.com/aws/aws-sdk-go-v2/blob/main/service/s3/api_client.go#L448
From there, the S3 client will be initialized with the desired configurations

Possible solutions to consider: [Functional Options Pattern](https://golang.cafe/blog/golang-functional-options-pattern.html)

hoangnv-bkhn:
## Description

- Problem: 
When the S3 client is initialized for retrievers and exporters, the application does not allow the user to further configure how the S3 client should behave. (It currently only accepts `AwsConfig` as a parameter).
- The way it was resolved:
Add an optional field to the Retriever/Exporter structure
Use this field when initializing the S3 client
- How to test the change:
Refer to the options that AWS allows to configure [here](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/s3#Options) and implement the appropriate custom functions.

## Closes issue(s)

Resolve #2451 

## Checklist
- [x] I have tested this code


thomaspoignant (approved):
Thanks for this addition, this is really great to have a more extensible s3 integration.

PR #2482, opened 2024-10-08
fix(tracing): Restore OTel tracing support

hairyhenderson:
## Description

In https://github.com/thomaspoignant/go-feature-flag/pull/1503, the code that used to call `s.otelService.Init` was accidentally removed. This restores that code.

Here's proof of spans once again being sent to a local Alloy/Tempo setup:

image

Note that I _have not_ written a unit test to cover this, as it's quite difficult to do so with how the OTel SDK is configured currently.

I have some ideas on how tracing support can be made more flexible (through use of `autoexport` primarily, and some other changes). I'll issue a PR for that some other time.

## Closes issue(s)

Fixes #2479 

## Checklist
- [x] I have tested this code
- [ ] I have added unit test to cover this code
- [ ] I have updated the documentation (`README.md` and `/website/docs`)
- [x] I have followed the [contributing guide](CONTRIBUTING.md)

Issue #2479, opened 2024-10-07
(bug) Tracing no longer working


hairyhenderson:
### Observed behaviour

Since https://github.com/thomaspoignant/go-feature-flag/pull/1503, the OTel service's `Init()` method is no longer being called. As a result, tracing is no longer working.

### Expected Behavior

Tracing should work!

hairyhenderson:
I've got a local fix in play, PR coming soon...

thomaspoignant:
Hey @hairyhenderson I am sorry for that, I did not spotted this regression.
If you have a PR ready can you open it and we can work from there.

hairyhenderson:
I've got a fix up here: https://github.com/thomaspoignant/go-feature-flag/pull/2482

PR #2485, opened 2024-10-09
feat(tracing): Misc. OTel tracing improvements

hairyhenderson:
## Description

I ran into a few papercuts while setting up tracing with the relay proxy after getting it up and running with #2482, most notably I use OTLP over gRPC rather than the default `http/protobuf` protocol.

So, this PR adds a few improvements:
- ability to disable with `OTEL_SDK_DISABLED=true`
- set endpoint with `OTEL_EXPORTER_OTLP_ENDPOINT` rather than config flag (but config flag is still supported)
- set protocol with `OTEL_EXPORTER_OTLP_PROTOCOL` (to `grpc`, for example)
- override service name with `OTEL_SERVICE_NAME`
- adds more resource attributes by default, including supporting setting [from the `OTEL_RESOURCE_ATTRIBUTES` env var](https://opentelemetry.io/docs/specs/otel/resource/sdk/#specifying-resource-information-via-an-environment-variable)
- adds support for configuring sampling strategy, including with the [Jaeger remote sampling protocol](https://pkg.go.dev/go.opentelemetry.io/contrib/samplers/jaegerremote#section-readme)
- hooks up the zap logger to OTel's error handling so log messages are formatted nicely, rather than just being logged with the stdlib `log.Print`

This also fixes a bug where `apiServer.Stop()` was not properly called. I've also modified the signature to receive a context so we can give the shutdown a 5s timeout.

## Closes issue(s)

None directly, but related to #2479

## Checklist
- [x] I have tested this code


thomaspoignant:
Hey @hairyhenderson thanks a lot for this PR.
I still need to look at it more in details, but I am not a big fan of using `os.GetEnv` outside of the configuration object.

I would prefer to have all those options in the [`Config`](https://github.com/thomaspoignant/go-feature-flag/blob/ff3d6069c73c60e9e8e494fad6709a3eb94b3dbd/cmd/relayproxy/config/config.go#L142) object rather than dealing with env variable directly.

We will still be able to set them via environment variable because we are using [`knadh/koanf`](https://github.com/knadh/koanf) that will allow us to set the values through file and/or environment variable to.

thomaspoignant:
I've played around with `koanf` and we could have something like this in the config:


```go
type Config struct {
// ...
	OtelConfig OpenTelemetryConfiguration `mapstructure:"otel" koanf:"otel"`
}


type OpenTelemetryConfiguration struct {
	SDK struct {
		Disabled bool `mapstructure:"disabled" koanf:"disabled"`
	} `mapstructure:"sdk" koanf:"sdk"`
	Exporter struct {
		Otlp struct {
			Endpoint string `mapstructure:"endpoint" koanf:"endpoint"`
			Protocol string `mapstructure:"protocol" koanf:"protocol"`
		} `mapstructure:"otlp" koanf:"otlp"`
	} `mapstructure:"exporter" koanf:"exporter"`
	Service struct {
		Name string `mapstructure:"name" koanf:"name"`
	} `mapstructure:"service" koanf:"service"`
	Traces struct {
		Sampler string `mapstructure:"sampler" koanf:"sampler"`
	} `mapstructure:"sampler" koanf:"sampler"`
	Resource struct {
		Attributes map[string]string `mapstructure:"attributes" koanf:"attributes"`
	} `mapstructure:"resource" koanf:"resource"`
}
```

This will allow to continue to set all the configuration through environment variable like you do, but at the same time be able to configure directly inside the configuration file also.

What do you think about this idea @hairyhenderson ? 
I am happy to help with that if you want.

hairyhenderson:
> What do you think about this idea @hairyhenderson ?

I'm OK either way... It's somewhat more common (in my experience) to see OTel configured through environment variables, but I understand your desire for both modes.

I'm not familiar with `koanf` so integrating this will take me a bit of time, and I don't have a lot of spare time to spend on this. I'll try to find some time over the next little while to get this updated...

thomaspoignant:
@hairyhenderson I can do it if you prefer. I am happy to update your PR with the changes if you don't have time.

thomaspoignant (approved):
@hairyhenderson I have refactored the configuration part.
We can now merge this PR and it will be available in the next version of GO Feature Flag.

hairyhenderson:
Thanks for this @thomaspoignant!! And sorry I wasn't able to make the changes, things have been busy for me lately 😅

thomaspoignant:
@hairyhenderson no worries.
Thank you for initiating it.

PR #2498, opened 2024-10-11
fix: file exporter path handling

### LINKED ISSUE #2483: (change) Improve Configuration of File Exporter
### Motivation

There are several issues with the `OutputDir` field in the configuration of the exporter that need improvement:
- `OutputDir` only accept existing path
Should we check the user-defined path, and create it if it doesn't exist, before trying to save variation events?

- Unexpected behavior when `OutputDir` is an empty string
Because the file path then looks like `/flag-variation-d171b59fa78c-1728460767.json`, which would cause it to be treated as an absolute path (instead of a relative path to the current working directory of the process) 
-> This will cause a `permission denied` error.
(Because the file is being created in the root directory (/) and on most systems, especially Unix-like systems (Linux, macOS), writing to the root directory requires elevated permissions (e.g., root user privileges)
The file path in this case should have a value of `flag-variation-d171b59fa78c-1728460767.json`

- Trailing Slash in `OutputDir`
The documentation recommends that OutputDir should end with `/`.
However, the current implementation makes the path look like `{OutputDir}//flag-variation-d171b59fa78c-1728460767.json`, resulting in a redundant `/`.

### Requirements

- Non-Existent Path Handling
The exporter should check the validity of the path, if it does not exist it should be created before proceeding with the data saving.

- `OutputDir` with empty string handling
If the user sets `OutputDir` to an empty string, the location where the exported data will be saved is the current working directory.

- Trailing Slash in `OutputDir`
The implementation should handle trailing slashes correctly to avoid redundant /.

hoangnv-bkhn:
## Description

- Problem: 
  - `OutputDir` in file exporter does not accept non-existent path.
  - If OutputDir is an empty string, the file path will become /{file-name}.json, which would cause it to be treated as an absolute path (instead of a relative path to the current working directory of the process).
-> This will result in a `permission denied` error.
  - The documentation recommends that `OutputDir` should end with `/`.
  However, the current implementation makes the path look like `{OutputDir}//{file-name}.json`, resulting in a redundant `/`.

- The way it was resolved:
Check the validity of OutputDir and create if necessary

- How to test the change:
Configure OutputDir with cases: path does not exist, path with trailing /, path is empty string, and observe the results

## Closes issue(s)

Resolve #2483 

## Checklist
- [x] I have tested this code


hoangnv-bkhn:
To edit docs, besides changing the content of the `.md` files in `/website/docs`, what else do I need to do?
It seems that the docs on the website are not updated according to what I edited

exporter/fileexporter/exporter.go
```diff
@@ -72,7 +72,19 @@ func (f *Exporter) Export(_ context.Context, _ *fflog.FFLogger, featureEvents []
 		return err
 	}
 
-	filePath := f.OutputDir + "/" + filename
+	// Handle empty OutputDir and remove trailing slash
+	outputDir := strings.TrimRight(f.OutputDir, "/")
+
+	var filePath string
+	if outputDir == "" {
+		filePath = filename
```
thomaspoignant, line 80: I have added a test case for when we have no `outputDir`.

exporter/fileexporter/exporter.go
```diff
@@ -72,7 +72,19 @@ func (f *Exporter) Export(_ context.Context, _ *fflog.FFLogger, featureEvents []
...
+	} else {
+		// Ensure OutputDir exists or create it
+		if err := os.MkdirAll(outputDir, 0755); err != nil {
```
thomaspoignant, line 83: I have change the permission to `0o644` to be consistent with what we are doing for the files.

thomaspoignant (approved):
@hoangnv-bkhn Thanks for your pull request, this makes the file exporter way better than it was before 🙇.

I have changed 2 NIT things:
- Change the permissions when creating the directory to `0o644` to align with the file creation.
- Added a test in the case we have no `outputDir`.

Everything else is really great.
I will merge it soon and it will be available in the next release.

thomaspoignant:
> It seems that the docs on the website are not updated according to what I edited

Oh sorry I forgot to address this one.
The doc is versioned this is why you don't see the change, but if you add `/next` after `/docs` you are able to see the next version of the doc after the release _(ex: https://gofeatureflag.org/docs/next)._

In the preview for this PR it will be in https://deploy-preview-2498--go-feature-flag-doc-preview.netlify.app/docs/next

PR #2534, opened 2024-10-17
feat: Add config to allow surpressing notification on launch (flag cache load)

### LINKED ISSUE #2532: (change) Don't send "new flag added" notification upon `ffClient` initialization
## Problem 
The current code calls `retrieveFlagsAndUpdateCache` to "initialize" the flag cache upon ffClient initialization, then starts the FlagUpdaterDaemon to continuously monitor the changes.
ref: https://github.com/thomaspoignant/go-feature-flag/blob/9ce4b9354baa944924671cfb2dd6cc22fddda044/feature_flag.go#L105

Unfortunately, any calls to `retrieveFlagsAndUpdateCache` will unconditionally call `notifier.Notify` if there is a difference between the new cache and the old cache. Upon initialization, "old cache" is always an empty hash, thus always having a diff and and ALWAYS sending "new flag added" notification on every `ffClient` initialization.

This behavior can be quite annoying (or at least unnecessary) for most applications, as the "new flag added" notification always gets sent every time in any kind of rolling deploy (the container with the old code gets replaced by another container with the new code).

## How (I think) It should be

I think it's better to NOT call the notifier(s) on the first cache load, because at this moment of initialization, the 'diff in cache' being detected are almost always  due to previous cache being empty. 

At least, I think there should be an option to skip the initial notification upon cache initialization.

## Note

I can prepare a PR if  okay with the proposal.

### Requirements

* Don't send notification when the cache diff is detected due to cache initialization

8ma10s:
## Description

This pull request introduces a new feature to disable notifications on initialization and refactors the caching mechanism to support this feature. It also includes updates to the test suite to ensure the new functionality works as expected.

Problem
* go-feature-flag always sent notification of diff to all Notifierson ffClient initialization because it detects the difference between the initial cache (= always empty) and the flags retrieved.

Resolution
* introduce `DisableNotificationOnInit` configuration key. If set to true, avoid sending the notification on cache initialization (it will still send notification when the change is detected AFTER initialization)
  * Since the behavior (should) stays the same if the new config is NOT set, it should not introduce any breaking change.

How to test
* I've added some tests to make sure the behavior stays the same when the new config is not explicitly set. You can run `go test ./...`
* Or, simply initialize `ffClient` should do the trick too.

## Closes issue(s)

Resolve #2532 

## Checklist
- [x] I have tested this code


8ma10s:
Sorry about the repeated commits. I think I've fixed the issue, and hopefully it works now.

cmd/relayproxy/config/config.go
```diff
@@ -175,6 +175,14 @@ type Config struct {
 	// Default: false
 	EnablePollingJitter bool `mapstructure:"enablePollingJitter" koanf:"enablepollingjitter"`
 
+	// DisableNotificationOnInit (optional) set to true if you do not want to
+	// send a notification when the flags are loaded.
+	// This is useful if you do not want a Slack/Webhook notification saying that
+	// the flags have been added every time you start the application.
+	// Default is set to false for backward compatibility.
+	// Default: false
+	DisableNotificationOnInit bool `mapstructure:"disableNotificationOnInit" koanf:"disablenotificationoninit"`
```
thomaspoignant, line 184: While I understand why you've put `notification` here, but I think it will be more explicit to use the name `notifier` which is what we are using everywhere.


```suggestion
	// DisableNotifierOnInit (optional) set to true if you do not want to
	// send a notification when the flags are loaded.
	// This is useful if you do not want a Slack/Webhook notification saying that
	// the flags have been added every time you start the application.
	// Default is set to false for backward compatibility.
	// Default: false
	DisableNotifierOnInit bool `mapstructure:"disableNotificationOnInit" koanf:"disablenotificationoninit"`
```

config.go
```diff
@@ -28,6 +28,14 @@ type Config struct {
 	// Default: false
 	EnablePollingJitter bool
 
+	// DisableNotificationOnInit (optional) set to true if you do not want to send
+	// a notification when the flags are loaded.
+	// This is useful if you do not want a Slack/Webhook notification saying that
+	// the flags have been added every time you start the application.
+	// Default is set to false for backward compatibility.
+	// Default: false
+	DisableNotificationOnInit bool
```
thomaspoignant, line 37: Same as before

```suggestion
	// DisableNotifierOnInit (optional) set to true if you do not want to send
	// a notification when the flags are loaded.
	// This is useful if you do not want a Slack/Webhook notification saying that
	// the flags have been added every time you start the application.
	// Default is set to false for backward compatibility.
	// Default: false
	DisableNotifierOnInit bool
```

feature_flag.go
```diff
@@ -203,8 +203,11 @@ func (g *GoFeatureFlag) startFlagUpdaterDaemon() {
 	}
 }
 
-// retrieveFlagsAndUpdateCache is called every X seconds to refresh the cache flag.
-func retrieveFlagsAndUpdateCache(config Config, cache cache.Manager, retrieverManager *retriever.Manager) error {
+func retreiveFlags(
```
thomaspoignant, line 206: I think this function deserve a comment.

feature_flag.go
```diff
@@ -262,8 +265,38 @@ func retrieveFlagsAndUpdateCache(config Config, cache cache.Manager, retrieverMa
 			newFlags[flagName] = value
 		}
 	}
+	return newFlags, nil
+}
+
+// retrieveFlagsAndInitializeCache is called when the feature flag is initialized
+func retrieveFlagsAndInitializeCache(config Config, cache cache.Manager, retrieverManager *retriever.Manager) error {
```
thomaspoignant, line 272: Rather having almost the same function for `retrieveFlagsAndInitializeCache` and `retrieveFlagsAndUpdateCache` I would have prefer to add a parameter to the previous function `retrieveFlagsAndUpdateCache` and use the same code on both side.

Something like 
```go
func retrieveFlagsAndUpdateCache(config Config, cache cache.Manager, retrieverManager *retriever.Manager, isInitialization bool) error {

// ...

    if isInitialization && config.DisableNotificationOnInit {
		err = cache.UpdateCache(newFlags, config.internalLogger)
	} else {
		err = cache.UpdateCacheAndNotify(newFlags, config.internalLogger)
	}

```

internal/cache/cache_manager.go
```diff
@@ -61,6 +62,14 @@ func (c *cacheManagerImpl) ConvertToFlagStruct(loadedFlags []byte, fileFormat st
 }
 
 func (c *cacheManagerImpl) UpdateCache(newFlags map[string]dto.DTO, log *fflog.FFLogger) error {
+	return c.updateCache(newFlags, log, false)
+}
+
+func (c *cacheManagerImpl) UpdateCacheAndNotify(newFlags map[string]dto.DTO, log *fflog.FFLogger) error {
+	return c.updateCache(newFlags, log, true)
+}
```
thomaspoignant, line 70: Do we need those functions or should we just keep `UpdateCache` with the new parameter?

thomaspoignant (commented):
Thanks a lot for this PR @8ma10s this is really great.

I like your approach but I will tend to simplify it a bit by just adding some parameters rather than creating new functions that can be hard to understand what is happening.
Would it work for you too?

internal/cache/cache_manager.go
8ma10s: ### Conclusion

Sure, I think I can change the code to do that. Let me change my code.
Given that it's an internal interface that's not meant to be used by anything outside of this repo's code, I think we don't need to care so much for readability (or method clarity) as much as I initially thought.

### My initial thought

I just thought that, given that `CacheManager` is (technically) defined as an interface, it would be beneficial to have a separate method to clearly indicate what the method is doing on the inside, rather than having an argument that doesn't really convey anything to the reader of the code.

What I mean is, if I have a code that calls this method defined in the interface, it would be (slightly) easier for the readers to understand what's going on when written like
```go
err = cache.UpdateCacheAndNotify(newFlags, config.internalLogger)
// do other stuff...
```
rather than
```go
err = cache.UpdateCache(newFlags, config.internalLogger, true)
// do other stuff...
```
because the value of `true` doesn't really convey anything just by reading it, whereas the method name `UpdateCacheAndNotify` clearly tells that it's gonna "notify something".

cmd/relayproxy/config/config.go
8ma10s: I think it makes sense, thank you. Yes, "notification" does seem a bit vague (as there can be other kinds of notifications), but "notifier" in this repo's context is pretty clear. Let me change that 👍

feature_flag.go
8ma10s: Same deal as here: https://github.com/thomaspoignant/go-feature-flag/pull/2534#discussion_r1811545406

I think I just tend to prefer having explicit function/method names just to make it a bit more readable on the caller code and avoid introducing too many parameters in a single method, but I think either works. To make it consistent with the code on the above comment, let me change this one as well 👍

feature_flag.go
```diff
@@ -262,8 +267,18 @@ func retrieveFlagsAndUpdateCache(config Config, cache cache.Manager, retrieverMa
...
+// retrieveFlagsAndUpdateCache is a function that retrieves the flags from the retrievers,
+// and update the cache with the new flags.
+func retrieveFlagsAndUpdateCache(config Config, cache cache.Manager, retrieverManager *retriever.Manager, isInit bool) error {
+	newFlags, err := retreiveFlags(config, cache, retrieverManager)
+	if err != nil {
+		return err
+	}
 
-	err := cache.UpdateCache(newFlags, config.internalLogger)
+	err = cache.UpdateCache(newFlags, config.internalLogger, isInit && !config.DisableNotifierOnInit)
```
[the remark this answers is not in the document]
thomaspoignant: Inverting the logic is something I am open with but we will need to explicitly guide GOFF users about this change in the future.

I'll keep it in mind in the potential breaking changes for the future.

feature_flag_test.go
```diff
@@ -706,3 +708,71 @@ func Test_UseCustomBucketingKey(t *testing.T) {
...
+	defer m.mu.Unlock()
+	m.notifyCalled = true
+	return nil
+}
+
+func (m *mockNotificationService) wasNotifyCalled() bool {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	return m.notifyCalled
+}
```
thomaspoignant, line 778: There is already a mock notifier available in the [`testutils/mock` directory](https://github.com/thomaspoignant/go-feature-flag/blob/main/testutils/mock/notifier_mock.go).

I think it would be better to use this one instead.

internal/cache/cache_manager_test.go
```diff
@@ -499,11 +499,152 @@ func Test_persistCacheAndRestartCacheWithIt(t *testing.T) {
...
+
+type mockNotificationService struct {
+	notifyCalled int
+}
+
+func (m *mockNotificationService) Notify(oldFlags, newFlags map[string]flag.Flag, logger *fflog.FFLogger) {
+	m.notifyCalled++
+}
+
+func (m *mockNotificationService) Close() {}
```
thomaspoignant, line 650: Same as the other test, I think we could use the already existing mock notifier available in the [`testutils/mock` directory](https://github.com/thomaspoignant/go-feature-flag/blob/main/testutils/mock/notifier_mock.go).

variation_test.go
```diff
@@ -46,7 +46,10 @@ func (c *cacheMock) GetLatestUpdateDate() time.Time {
 func (c *cacheMock) ConvertToFlagStruct(_ []byte, _ string) (map[string]dto.DTO, error) {
 	return nil, nil
 }
-func (c *cacheMock) UpdateCache(_ map[string]dto.DTO, _ *fflog.FFLogger) error {
+func (c *cacheMock) UpdateCache(_ map[string]dto.DTO, _ *fflog.FFLogger, _ bool) error {
+	return nil
+}
+func (c *cacheMock) UpdateCacheAndNotify(_ map[string]dto.DTO, _ *fflog.FFLogger) error {
 	return nil
 }
```
thomaspoignant, line 54: ✂️ This function is not part of the interface anymore, we should probably remove it from the mock.

website/docs/go_module/configuration.md
```diff
@@ -24,6 +24,7 @@ During the initialization you must give a [`ffclient.Config{}`](https://pkg.go.d
...
+| `DisableNotifierOnInit`       | *(optional)* If **true**, the SDK will not send a notification when the flags are loaded during initialization. This is useful if you do not want a Slack/Webhook notification saying that the flags have been added every time you start the application.<br/>Default: **false**                                                                                                                                                                                                                                                                                                                                                                                                                              |
```
thomaspoignant, line 27:
```suggestion
| `DisableNotifierOnInit`       | *(optional)* If **true**, the SDK will not call any notifier when the flags are loaded during initialization. This is useful if you do not want a Slack/Webhook notification saying that the flags have been added every time you start the application.<br/>Default: **false**                                                                                                                                                                                                                                                                                                                                                                                                                              |
```

website/docs/relay_proxy/configure_relay_proxy.md
```diff
@@ -38,6 +38,7 @@ ex: `AUTHORIZEDKEYS_EVALUATION=my-first-key,my-second-key`)_.
...
 | `enablePollingJitter`             | boolean                                | `false`            | Set to true if you want to avoid having true periodicity when retrieving your flags. It is useful to avoid having spike on your flag configuration storage in case your application is starting multiple instance at the same time.<br/>We ensure a deviation that is maximum ±10% of your polling interval.<br />Default: false                                                                                                                 |
+| `DisableNotifierOnInit`       | boolean                                | `false`            | If **true**, the relay proxy will not send a notification when the flags are loaded during initialization. This is useful if you do not want a Slack/Webhook notification saying that the flags have been added every time you start the proxy.<br/>Default: **false**                                                                                                                                                                                                         |
```
thomaspoignant, line 41:
```suggestion
| `DisableNotifierOnInit`       | boolean                                | `false`            | If **true**, the relay proxy will not call any notifier when the flags are loaded during initialization. This is useful if you do not want a Slack/Webhook notification saying that the flags have been added every time you start the proxy.<br/>Default: **false**                                                                                                                                                                                                         |
```

thomaspoignant (changes requested):
Thanks again, we are almost ready to go and to merge this PR.

I have just put some comments on the mocks because we can probably reuse the existing one instead of creating new ones.

feature_flag_test.go
8ma10s: Oops, I didn't notice that. Fixed!
(I've also added the missing `mutex` logic to the Notifier mock, since it seems like other mocks like `exporter_mock` have similar mutex control to avoid race condition)

internal/cache/cache_manager_test.go
8ma10s: Actually, the existing Notifier mock could not be used because:
* input parameter for `Notify()` are different
* doesn't implement `Close()` method.

However, I think it makes sense to move the mock to `testutils/mock` to be more consistent.
So I've created a separate mock in `testutils/mock` directory and used the new mock.

thomaspoignant (approved):

8ma10s:
@thomaspoignant Thank you for reviewing and giving me suggestions! 🚀

PR #2535, opened 2024-10-17
feat: support for multiple exporters

### LINKED ISSUE #2505: (feature) Support for multiple exporters
### Requirements

The addition of support for multiple exporters increases the efficiency of the data export feature.

Some of the benefits include:
- Redundancy and Reliability:
  - Backup: If one exporter fails (e.g., an HTTP endpoint is down), having another exporter (e.g., a file exporter) ensures that the data is still captured.
- Different Use Cases:
  - Auditing: There may be a need to export data to secure, immutable storage for audit purposes.
  - Monitoring: Simultaneously, we might want to send data to a monitoring system, such as a logging service, for real-time analysis.
  - Analytics: Exporting data to an analytics platform can help in understanding user behavior and feature adoption.
- Separation of Concerns:
  - Security: Sensitive data might be exported to a secure location, while less sensitive data can be sent to a more accessible endpoint.
  - Performance: Exporting to a local file might be faster and less resource-intensive, while exporting to a remote server might be more useful for centralized data collection.

hoangnv-bkhn:
## Description

- Support for configuring multiple exporters simultaneously
- The way it was resolved:
Each data exporter will be initialized with a separate schedule to operate independently of each other.
- How to test the change:
Refer: [Example](https://github.com/hoangnv-bkhn/go-feature-flag/blob/feat/support-for-multiple-exporters/examples/data_export_log_and_file/main.go)

## Closes issue(s)

Resolve #2505 

## Checklist
- [x] I have tested this code


.github/workflows/unassign-issue.yml
```diff
@@ -107,8 +107,7 @@ jobs:
             const staleIssues = await github.paginate(github.rest.issues.listForRepo, {
               owner, repo,
               state: 'open',
-              labels: staleLabel,
-              assignee: '*'
```
hoangnv-bkhn, line 111: I modified this to handle the assignee unassigning themselves (while the `Assigned` label remains)

thomaspoignant: can we do this in another PR to avoid mixing things?

thomaspoignant:
@hoangnv-bkhn I haven't looked in detail to all the PR but there is something I would have done differently.

In your proposal you are duplicating the logic of the `DataExporter` by allowing users to have a different different `dataExporterSchedulers` that allows to set different flush interval and max buffer per `exporter`.

While I understand why you've done that I think it will be easier to have only 1 flush interval for all the exporters.
Have you consider adding a field `exporters` in the [`DataExporter`](https://github.com/thomaspoignant/go-feature-flag/blob/main/config_collector.go) struct?

Any reason why you have not followed that path?

hoangnv-bkhn:
> @hoangnv-bkhn I haven't looked in detail to all the PR but there is something I would have done differently.
> 
> In your proposal you are duplicating the logic of the `DataExporter` by allowing users to have a different different `dataExporterSchedulers` that allows to set different flush interval and max buffer per `exporter`.
> 
> While I understand why you've done that I think it will be easier to have only 1 flush interval for all the exporters. Have you consider adding a field `exporters` in the [`DataExporter`](https://github.com/thomaspoignant/go-feature-flag/blob/main/config_collector.go) struct?
> 
> Any reason why you have not followed that path?

@thomaspoignant 
Actually, that was the first way I thought of and implemented.
But then I realized that users might need to set different values for flush interval and max buffer for different exporters.
So I changed it this way to give users maximum configurability.

.github/workflows/unassign-issue.yml
hoangnv-bkhn: Okay, that makes sense.

hoangnv-bkhn:
@thomaspoignant 
[Kindly Reminder]
What do you think of my idea?

If you think that the idea of all exporters having the same flush interval and max buffer configuration makes more sense at the moment, then we can reconsider it.

thomaspoignant:
@hoangnv-bkhn Sorry I am lagging with the reviews those days.
I understand why you prefer this option and I tend to agree that this is a good idea.

My main concern with your solution is that you add events to each exporters which means that we duplicate the usage of memory by the number of exporters and I want to avoid that as much as possible.
We can maybe have a common cache for every exporters, but this may be tricky to handle failures etc...

I am still trying to understand what is the best solution here, but I am happy to get your inputs.

hoangnv-bkhn:
@thomaspoignant 
In case you missed my message, I messaged you on Slack Channel
Would love to hear your thoughts.

hoangnv-bkhn:
Hi @thomaspoignant :wave:
Sorry for the delay, 
I just updated the handling of the multiple exporters.
Now they can work independently using the shared cache for better memory management.
Please help me review this PR at your convenience! 🙇

thomaspoignant:
Hello @hoangnv-bkhn and sorry for the long wait for this one.
I have open a new PR #3171 inspired by your PR to support multiple exporters.

The main difference is that the exporter is now using generics in order to be prepared for `tracking` events that will come in the future.

thomaspoignant:
Hello @hoangnv-bkhn and sorry for the long wait for this one.
I have open a new PR #3171 inspired by your PR to support multiple exporters.

The main difference is that the exporter is now using generics in order to be prepared for `tracking` events that will come in the future.
I will close this PR and release the other one, let me know if it solves your issue.

hoangnv-bkhn:
Hello @thomaspoignant, 
Thank you for your work on the feature. I'm glad to see that new feature has been released. It's a valuable improvement. So no problem, it's a learning experience for me as well, and I appreciate the explanation of the changes. 🙇

PR #2563, opened 2024-10-23
feat: Detect undefined variations

thomaspoignant:
## Description
In this PR we had a new check when validating a feature flag to check if we reference an undefined variation in one of the rule configuration.

The effect of this PR is double, it will allow the linter to detect those issues and it will discard this flag (and log an error)  for misconfigured flags.

## Closes issue(s)
Resolve #2652

## Checklist
- [x] I have tested this code

### DIFF
--- a/internal/flag/rule.go
+++ b/internal/flag/rule.go
@@ -223,7 +223,7 @@ func (r *Rule) MergeRules(updatedRule Rule) {
 }
 
 // IsValid is checking if the rule is valid
-func (r *Rule) IsValid(defaultRule bool) error {
+func (r *Rule) IsValid(defaultRule bool, variations map[string]*interface{}) error {
 	if !defaultRule && r.IsDisable() {
 		return nil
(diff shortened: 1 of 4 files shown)


internal/flag/rule.go
```diff
@@ -240,8 +240,11 @@ func (r *Rule) IsValid(defaultRule bool) error {
 	// Validate the percentage of the rule
 	if r.Percentages != nil {
 		count := float64(0)
-		for _, p := range r.GetPercentages() {
+		for k, p := range r.GetPercentages() {
 			count += p
+			if _, ok := variations[k]; !ok {
+				return fmt.Errorf("invalid percentage: variation %s does not exists", k)
```
kna, line 246: It might be better to change like this?
```suggestion
				return fmt.Errorf("invalid percentage: variation %s does not exist", k)
```

kna:
Thanks!

PR #2569, opened 2024-10-24
fix: Notifier should fire during runtime (if isInit is false)

8ma10s:
## Description
This is a bug fix for #2534 

#2534 had an incorrect logic of NOT calling notifier when the status is not set to `isInit` (=upon polling).
It should:
* NOT call notifier only if `isInit`  && `DisableNotifierOnInit` are BOTH true
* call the notifier in any other cases

### DIFF
--- a/feature_flag.go
+++ b/feature_flag.go
@@ -279,7 +279,7 @@ func retrieveFlagsAndUpdateCache(config Config, cache cache.Manager,
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


8ma10s:
@thomaspoignant sorry about that. I think the code currently doesn't call the notifiers if `isInit` is false, which probably affects all users.

As a hotfix, I've created a PR that does just the fix of the logic.

thomaspoignant (approved):
oh good catch, I missed this one.
Let me release the fix.

PR #2579, opened 2024-10-26
fix(otel): Don't require entrypoint config to initialize OTel

### LINKED ISSUE #2578: (bug) Tracing only initializes when specifying endpoint through config
### Observed behaviour

After #2482 and #2485, tracing won't initialize unless the `openTelemetryOtlpEndpoint` config entry is set. When I'm using `OTEL_EXPORTER_OTLP_ENDPOINT` instead I'd rather not have to specify it in the configuration as well.

### Expected Behavior

Tracing should be disabled via `OTEL_SDK_DISABLED=true`, otherwise it should be enabled by default.

### Steps to reproduce

_No response_

hairyhenderson:
## Description

Changes behaviour to always call `otelService.Init`, to avoid requiring setting the `openTelemetryOtlpEndpoint` config element.

If the OTel SDK is not explicitly disabled and no OTel endpoint is listening, the OTel SDK will complain in the logs. Since I don't want to unnecessarily spam logs, I'm adding support for the [`OTEL_LOG_LEVEL`](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration) variable, but defaulting to `debug` rather than the spec's `info` for somewhat more sane logs.

Notes: I've not done it here, but I think deprecating `openTelemetryOtlpEndpoint` in the future is probably a good idea, in preference for `otel.exporter.otlp.endpoint`. I've also noticed that the `otel` config section is undocumented - I may have a look at that in a separate PR.

## Closes issue(s)

Fixes #2578 

## Checklist
- [x] I have tested this code


thomaspoignant (commented):
@hairyhenderson  I have tested this new version and my main concern is that we are still trying to call the OTLP when nothing is configured.

We have this log for every query:
```
{"level":"debug","ts":1730195483.37956,"caller":"opentelemetry/otel.go:93","msg":"OTel error","error":"traces export: Post \"https://localhost:4318/v1/traces\": dial tcp [::1]:4318: connect: connection refused"}
```
Maybe we should set the `noop.NewTracerProvider()` when we have no OTLP endpoint configured.

hairyhenderson:
> We have this log for every query:

yeah, that's why I set it to log at debug level...

> Maybe we should set the `noop.NewTracerProvider()` when we have no OTLP endpoint configured.

Yeah, the issue is that if there's no exporter, defaults will be used (OTLP over http/protobuf on https://localhost:4318). I had thought that perhaps some users would want this. 

Realistically though, I think I'm the only user of GOFF that needs the OTel support so far, so requiring the endpoint to be set is probably fine.

I'll update this to check `openTelemetryOtlpEndpoint` or `otel.exporter.otlp.endpoint` for now.

hairyhenderson:
@thomaspoignant I've simplified things - PTAL 🙂

thomaspoignant (approved):
Looks really good now 👍

PR #2609, opened 2024-11-02
feat: Cmd evaluation

### LINKED ISSUE #2560: (feature) evaluation command line for local testing
### Context
When using GO Feature Flag in case of percentage based evaluation sometimes you want to know which variant will be returned for a specific evaluation context.

### Proposed solution
We want a command line tool that allow to perform the evaluation directly in the terminal allowing to get a fast feedback on which variant is served for a specific flag configuration file.

#### Example

```shell
$ go-feature-flag-cli evaluate --config ./flag.goff.yaml --flag "my-flag-name" --evaluation-context '{"targetingKey": "03b77ca4-b599-4a43-8241-af67b23b8fc5"}'

{
  "trackEvents": false,
  "variationType": "Variation_1",
  "failed": false,
  "version": "",
  "reason": "STATIC",
  "errorCode": "",
  "value": true,
  "cacheable": true
}

```

The cli should allow those commands:
- `-c` or `--config`: path to the flag configuration file **(mandatory)**
- `-f` or `--flag`: name of the flag to evaluate **(optional)**, if not set the cli should evaluate all the flags available in the `-f` file configuration.
- `-e` or `--evaluation-context`: the evaluation context to validate the flag against the file.

### Implementation
Here are some guidelines for the implementation:
- The code for this command line should be under `cmd/cli`.
- We can use a framework to manage this command line, my choice will be to use [`spf13/cobra`](https://github.com/spf13/cobra).
- Consider that the `evaluate` sub command will be the 1st one but we want to extend the usage of this command line to other need in the future _(maybe move the linter there also)._
- The logic behind this command line will be similar to the `editor` API available [here](https://github.com/thomaspoignant/go-feature-flag/tree/main/cmd/editor).

mexes20:
## Description

evaluation command line for local testing

## Closes issue #2560

Resolve  #2560

## Checklist
- [ ] I have tested this code


cmd/cli/lint.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func runLint(cmd *cobra.Command, args []string) error {
+	config, err := loadConfig(configFile)
+	if err != nil {
+		return err
+	}
+
+	result := lintConfig(config)
+	output, err := json.MarshalIndent(result, "", "  ")
```
thomaspoignant, line 37: **issue:** There is no import for json so the code is not compiling.

cmd/cli/lint.go
```diff
@@ -0,0 +1,87 @@
...
+	if err != nil {
+		return fmt.Errorf("error formatting output: %w", err)
+	}
+
+	fmt.Println(string(output))
+	return nil
+}
+
+func lintConfig(config *FlagConfig) *LintResult {
```
thomaspoignant, line 46: **suggestion:** You have re-implemented the lint logic that is already available in `cmd/lint/linter.go`, I would suggest to replace your logic by this one in order to stay consistent.

cmd/cli/lint.go
```diff
@@ -0,0 +1,87 @@
...
+			result.Warnings = append(result.Warnings, 
+				fmt.Sprintf("Flag '%s' has no default rule", name))
+		}
+	}
+
+	return result
+}
+
+func isValidFlagName(name string) bool {
+	return strings.ToLower(name) == name && !strings.Contains(name, " ")
```
thomaspoignant, line 86: **question:** I am not sure to get what you are testing here? There is no need to have a lower case name, and flag names can contains spaces.

cmd/cli/evaluate.go
```diff
@@ -0,0 +1,160 @@
...
+	if err := json.Unmarshal([]byte(contextJSON), &context); err != nil {
+		return nil, fmt.Errorf("error parsing evaluation context: %w", err)
+	}
+
+	targetingKey, ok := context["targetingKey"].(string)
+	if !ok {
+		return nil, fmt.Errorf("targetingKey must be a string")
+	}
+
+	user := ffuser.NewUser(targetingKey)
```
thomaspoignant, line 97: **suggestion:** `NewUser` is a deprecated function, can you use `ffcontext.NewEvaluationContext` instead?

```suggestion
	evalCtx := ffcontext.NewEvaluationContext(targetingKey)
```

cmd/cli/evaluate.go
```diff
@@ -0,0 +1,160 @@
...
+	for k, v := range context {
+		if k != "targetingKey" {
+			user.Custom[k] = v
```
thomaspoignant, line 100: **issue:** `Custom` is not a public attribute so you can't add things in the context like this.

**suggestion:** Use the function `AddCustomAttribute(k, v)` instead
```suggestion
			evalCtx.AddCustomAttribute(k, v)
```

cmd/cli/evaluate.go
```diff
@@ -0,0 +1,160 @@
...
+
+	var config FlagConfig
+	if err := yaml.Unmarshal(data, &config); err != nil {
+		return nil, fmt.Errorf("error parsing config file: %w", err)
+	}
+
+	return &config, nil
+}
+
+func parseContext(contextJSON string) (*ffuser.User, error) {
```
thomaspoignant, line 86: **suggestion:** `ffuser.User` is deprecated, please use `ffcontext.EvaluationContext` instead.

cmd/cli/evaluate.go
```diff
@@ -0,0 +1,160 @@
...
+type EvaluationResult struct {
+	TrackEvents   bool        `json:"trackEvents"`
+	VariationType string      `json:"variationType"`
+	Failed        bool        `json:"failed"`
+	Version       string      `json:"version"`
+	Reason        string      `json:"reason"`
+	ErrorCode     string      `json:"errorCode"`
+	Value         interface{} `json:"value"`
+	Cacheable     bool       `json:"cacheable"`
+}
```
thomaspoignant, line 25: **question:** Why do you need to create a new response struct here? You can probably use `model.VariationResult` instead that has the exact same signature _(except that it is a generic type)_.

This struct is available [here](https://github.com/thomaspoignant/go-feature-flag/blob/8f53d070de7ba82a626275348b1f57f2e353f2f5/model/variation_result.go#L11).

cmd/cli/evaluate.go
```diff
@@ -0,0 +1,160 @@
...
+	output, err := json.MarshalIndent(result, "", "  ")
+	if err != nil {
+		return fmt.Errorf("error formatting output: %w", err)
+	}
+
+	fmt.Println(string(output))
+	return nil
+}
+
+func evaluateSingleFlag(flag *variation.Flag, user *ffuser.User) (*EvaluationResult, error) {
```
thomaspoignant, line 127:
```suggestion
func evaluateSingleFlag(flag *variation.Flag, user *ffcontext.EvaluationContext) (*EvaluationResult, error) {
```

cmd/cli/evaluate.go
```diff
@@ -0,0 +1,160 @@
...
+		ErrorCode:     eval.ErrorCode,
+		Value:         eval.Value,
+		Cacheable:     eval.Cacheable,
+	}, nil
+}
+
+func evaluateAllFlags(config *FlagConfig, user *ffuser.User) error {
+	results := make(map[string]*EvaluationResult)
+
+	for flagName, flag := range config.Flags {
```
thomaspoignant, line 145: **issue:** Here `config.Flags` will always be empty since you've passed nothing for the flag name here.

**suggestion:** You can directly use the function [`AllFlagsState`](https://github.com/thomaspoignant/go-feature-flag/blob/8f53d070de7ba82a626275348b1f57f2e353f2f5/variation_all_flags.go#L11) that perform the evaluation for all the flags.

thomaspoignant (changes requested):
Hey @mexes20 thanks for your contribution, I am looking forward for this new cli.

I have made a first pass and found some things to fix before being able to have a working command line.
Can you try to fix it and ensure that the code compiles before re-requesting a review?

Thanks again for your contribution 🙏

thomaspoignant (approved):

PR #2672, opened 2024-11-16
feat: Azure Blob Storage Retriever

### LINKED ISSUE #2558: (feature) Azure Blob Storage retriever
### Requirements

In order to continue GO Feature Flag support to major cloud provider we want to bring the same support as GCP and AWS for Azure.

To do this, we want to let GO Feature Flag users the possibility to store their configuration files into Azure Blob Storage.

For this, we want to have a Azure Blob Storage `retriever` that will get the configuration directly from the Azure service.

**Things to take in consideration:**
- We can use the azure SDK to be able to communicate with BlobStorage (https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/storage/azblob)
- We should be able to give the credentials through the normal mechanism of authentifcation of the GO sdk.
- This should work both as GO module and in the relay-proxy.

_An example of PR to add a new retriver https://github.com/thomaspoignant/go-feature-flag/pull/1717/files_

Abeeujah:
## Description

Azure Blob Storage Retriever to get feature flag configurations directly from Azure.

## Closes issue(s)

Resolve #2558 

## Checklist
- [x] I have tested this code


thomaspoignant (changes requested):
@Abeeujah thanks for your PR, I've made some comments and the PR needs some changes.

Abeeujah:
Hi @thomaspoignant my bad; I didn't update the tests after implementing the `InitializableRetriever` Interface; it's been fixed now, can you review again? all tests are now passing...

PR #2708, opened 2024-11-22
feat(metrics): Collect process and Go runtime metrics

hairyhenderson:
## Description

A downside of using a custom Prometheus registry rather than the global one is that it has no collectors available. Generally it's quite useful to have the [Go collector](https://pkg.go.dev/github.com/prometheus/client_golang@v1.20.5/prometheus/collectors#NewGoCollector) as well as the [Process collector](https://pkg.go.dev/github.com/prometheus/client_golang@v1.20.5/prometheus/collectors#NewProcessCollector) enabled, to track memory, goroutine, file descriptor, CPU, and other metrics useful for observing any Go process.

This PR enables those two.

Before:
```console
$ curl -s http://localhost:1034/metrics | wc -l
29
$ curl -s http://localhost:1034/metrics | grep go_
$
```

After:
```console
$ curl -s http://localhost:1034/metrics | wc -l
122
$ curl -s http://localhost:1034/metrics | grep go_goroutines
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 13
$
```

## Closes issue(s)

None - small enough change I didn't think it necessary!

## Checklist
- [x] I have tested this code


thomaspoignant (approved):
Hey thanks for this one, I agree this is simple enough not to have any issue.
🚀

PR #2710, opened 2024-11-22
feat(pprof): Allow enabling pprof endpoint separately from debug flag

### LINKED ISSUE #2709: (feature) Should be possible to enable `/debug/pprof` without the additional logging
### Requirements

Right now it's necessary to set `debug: true` to get the pprof endpoint to listen. We like to continuously profile everything with Alloy and Pyroscope, but I'd rather not have requests to `/metrics`/`/health`/etc also be logged.

How about an `enablePprof` config item? I'll issue a quick PR for discussion, but thought you might have other ideas too...

hairyhenderson:
## Description

See #2709 

## Closes issue(s)

Resolve #2709

## Checklist
- [x] I have tested this code


thomaspoignant (approved):
Thanks for this PR @hairyhenderson.
I just added a bit of documentation but except that it is perfect 👌

PR #2712, opened 2024-11-22
feat(logging): Support logfmt

### LINKED ISSUE #2711: (feature) Logging in logfmt format as well as json
### Requirements

Right now, the relay proxy only logs in JSON format, which isn't as human-readable as logfmt. I suggest adding a new `logFormat` field to allow using logfmt.

I'll work on a PR for this.

hairyhenderson:
## Description

Adds logfmt support through a new `logFormat` global config setting. I ended up needing to refactor things a bit.

Also, the original `TestLoggerOutput` unit test actually wasn't testing the logger at all - that logger was being silently ignored. I fixed that test and merged it with the `TestInitLogger` test.

## Closes issue(s)

Resolve #2711 

## Checklist
- [x] I have tested this code


cmd/relayproxy/config/config.go
```diff
@@ -445,6 +451,10 @@ func (c *Config) IsValid() error {
 		}
 	}
 
+	if c.LogFormat != "" && c.LogFormat != "json" && c.LogFormat != "logfmt" {
+		return fmt.Errorf("invalid log format %s", c.LogFormat)
+	}
```
thomaspoignant, line 456: **💣 chore:** ‏I have refactor this part to use a switch and lowercase the value.

thomaspoignant (approved):
Super nice addition, it may be easier to read proper text logs than JSON you are right.

**🤓 nitpick:** ‏I've added a small test and refactor a bit the validation part.
**👏 praise:** ‏Thanks a lot also for taking the time to fix the `TestLoggerOutput`.

I will put everything in the `v1.39.0` version of GO Feature Flag.

hairyhenderson:
Thanks for the quick review @thomaspoignant!!

Issue #2314, opened 2024-09-05
(feature) Flexible support for many retrievers/retriever scanning in S3 bucket and subsequent collision handling


iisharankov:
### Requirements

We are a small team within a large company (~20k) aiming to deploy a FeatureFlag solution across the company. Our solution aims to centralize feature flagging between hundreds of teams, with about a dozen users each, where each team may have hundreds of flags over many different projects over several environments (eg: dev, stage, prod).

Our internal solution works, but Go-Feature-Flag is a promising replacement that offers to reduce our technical debt tremendously at our expense of several medium architectural changes to our side that we are happy to adopt. After some experimentation, we've found two simple but important hurdles for using go-feature-flag at larger scales as described. 

# Use case

Members in an arbitrary company team can create a project at any point for any tool within their domain of work. They expect to be able to add flags of any name (different projects in a team (or different teams) may therefore have the same flag name, but these are categorically different flags) to their projects and be able to access them through any client/server provider.

1. Each project has a unique UUID. 
2. Each project will always exist in a Dev, Stage, AND Prod environment (all three).
3. Each Dev, Stage, and Prod environment for a given project has a unique UUID.
4. The users have access to both these UUIDs from our front end to use in their providers.
5. In the back end all Dev, Stage, and Prod data exist in its respective AWS cluster in a S3 bucket.
6. The S3 bucket is shared by all projects in that environment, but each project has a special unique (hashed name) file that is generated automatically for that project, and all flags for that project exist in that unique file.

# Features
With this setup/scale, two limitations of go-feature-flag inhibit what feels could be a seamless integration into this system

## Accept multiple/all files in a given bucket (or folder in bucket)
Each time a user creates a project, a file is made for them to store their flags. This is done so no two projects have flags in a file together, both for security benefits and maintainability reasons. 

The current [s3 bucket retriever](https://gofeatureflag.org/docs/relay_proxy/configure_relay_proxy#s3) is designed for a single file within a bucket. For each new file, a new retriever needs to be added to the `values.yaml`. This is not ideal as it would mean adding 
```
retriever:
    kind: s3
    bucket: bucket-name
    item: new-project-file.yaml
```
to the relay-proxy configuration every time, and then propagating to dev/stage/prod and approving the MR.  This is difficult to automate and not very maintainable. A better solution would be to be able to provide a list of files such as 
```
retriever:
    kind: s3
    bucket: bucket-name
[...]

PR #2734, opened 2024-11-27
docs(specification): Flag set specification proposal

thomaspoignant:
This pull request contains a proposal specification for the flag set capability in GO Feature Flag (#2752).  
Based on recent discussions in the community, this is the most requested features we had recently.

In this PR there is a description of the problem space, plus some proposals on how it can look from a user perspective.
All feedbacks are more than welcome 🙏.

> [!NOTE]
>_If you want a more readable version of the document, go directly [here](https://github.com/thomaspoignant/go-feature-flag/blob/spec-flag-set/website/src/pages/specification/20241027-flagsets.md)._

---
For reference, a flag set is:

> A collection of related flags. This grouping helps organize feature flags based on their intended use, facilitating easier management and deployment.
_Source: [Openfeature glossary](https://openfeature.dev/specification/glossary/#flag-set)._


website/src/pages/specification/20241027-flagsets.md
```diff
@@ -0,0 +1,200 @@
...
+    path: config-file1.goff.yaml
+  - kind: file
+    path: config-file2.goff.yaml
+
+authorizedKeys:
+  evaluation:
+    - apikey1 # owner: userID1
+    - apikey2 # owner: userID2
+  admin:
+    - apikey3
```
thomaspoignant, line 138: 🤔 I am not sure how to configure the relation between the API keys and the flag sets, all proposals are welcome on this.

website/src/pages/specification/20241027-flagsets.md
```diff
@@ -0,0 +1,200 @@
...
+  - kind: file
+    path: config-file2.goff.yaml
+    flagSet: flagset-teamB
+
+authorizedKeys:
+  evaluation:
+    - apikey1 # owner: userID1
+    - apikey2 # owner: userID2
+  admin:
+    - apikey3
```
thomaspoignant, line 191: 🤔 I am not sure how to configure the relation between the API keys and the flag sets, all proposals are welcome on this.

website/src/pages/specification/20241027-flagsets.md
thomaspoignant: > I just went through the API keys configuration docs and it looks like it accept any string. Since it's an array it can be organized so that the available flag sets for a specific api key are declared as well in the config file.

Yes it can be anything here.
I am just not sure how to represent the configuration without breaking changes for existing config 🤔.

> Another approach would be to generate something like JWTs to be used as API keys and encode flag set access data as claims. This path would be much more complex
I was also thinking about that, but it could be an evolution for a later stage.

website/src/pages/specification/20241027-flagsets.md
```diff
@@ -0,0 +1,182 @@
...
+    variation: enabled
+```
+
+#### Relay-proxy configuration example
+```yaml
+# ...
+retrievers:
+  - kind: file
+    path: config-file1.goff.yaml
+    flagSet: flagset-teamA
```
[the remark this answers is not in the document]
iisharankov: I'm a little confused how this would work in practice though? If tomorrow a new team comes in and they need their own flag set, then GOFF would need to be redeployed again with a new configuration for another retriever. In our companies' deployment/FF solution, we're hoping each development project in each team is an individual flag set to avoid possible conflict, which means every new project on every team that's created would require a production redeployment of go-feature-flag on our side.  Obviously this isn't really maintainable, which is what our [feature request](https://github.com/thomaspoignant/go-feature-flag/issues/2314) aimed to solve.

Disregarding the feature request previously, would another solution such as having the `flagset` key defined per flag work? Each flag could have a `flagset` key that if not provided would be `default` (keeping original functionality), otherwise would set the flag as being owned by that `flagset`. The issue here is flags still have the same name in their source (S3, MongoDB, Redis, etc.), but this doesn't have to be an issue for goff as when it loads flags into its memory it can use the `flagset` to distinguish flags internally.

gogbog:
### Proposal:
 
**Solution**: In this solution, we configure each flag individually to determine whether it should be included in a set. This approach allows us to maintain a single retriever in the GOFF configuration with multiple FlagSets. Whenever a new flag is introduced, we can simply check the parameters in the JSON object to detect whether the user intends to include it in a FlagSet.

 Example:
```
featureA-enabled:
    flagSet: flagset-teamA
    variations:
        enabled: true
        disabled: false
    defaultRule:
        variation: enabled
```
 
**PROS**:
 
- My opinion is that this would be the most flexible solution, as the FlagSets will become highly dynamic and specific to individual flags.
- We will be able to control the FlagSets, without redeploying GOFF or adding new receivers.
- MongoDB would still work, as the best design would currently involve placing each flag within a single document (single JSON object).
 
**CONS**:
 
- This could be the hardest solution to implement yet, I'm not sure.
- Harder to keep up which flag is in which set
- Every time when a new flag gets introduced we would have to check if a new FlagSet comes up and modify the cache accordingly

thomaspoignant:
@gogbog your proposal was one of my original idea but we will not be able to have this statement:
> We want to allow 2 flags with the same name if they are used by different teams or platforms.

**Why ?**
Because the YAML file will be invalid if we try to redefine 2 flags with the same name.

**Example:**
```
featureA-enabled:
    flagSet: flagset-teamA
    variations:
        enabled: true
        disabled: false
    defaultRule:
        variation: enabled

featureA-enabled:
    flagSet: flagset-teamB
    variations:
        enabled: true
        disabled: false
    defaultRule:
        variation: enabled
```

Also it does not really allow a team to edit their file autonomously without impacting each others.

website/src/pages/specification/20241027-flagsets.md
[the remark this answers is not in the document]
thomaspoignant: >I'm a little confused how this would work in practice though? If tomorrow a new team comes in and they need their own flag set, then GOFF would need to be redeployed again with a new configuration for another retriever. In our companies' deployment/FF solution, we're hoping each development project in each team is an individual flag set to avoid possible conflict, which means every new project on every team that's created would require a production redeployment of go-feature-flag on our side. Obviously this isn't really maintainable, which is what our https://github.com/thomaspoignant/go-feature-flag/issues/2314 aimed to solve.

@iisharankov your use case is still in my mind but I want to first introduce the notion of flag set and later address your usecase.
All the proposals are still valid with your use case, it will just need to have some kind of meta retriever that will be able to automatically create a flag set.

I place your need for now in the **Out of scope** section, but as soon as we have the flag set ready we will be able to work on this too.

> Disregarding the feature request previously, would another solution such as having the flagset key defined per flag work? Each flag could have a flagset key that if not provided would be default (keeping original functionality), otherwise would set the flag as being owned by that flagset. The issue here is flags still have the same name in their source (S3, MongoDB, Redis, etc.), but this doesn't have to be an issue for goff as when it loads flags into its memory it can use the flagset to distinguish flags internally.

Yes that was my original idea, but we will have limits because of unicity of flag name here.
See the example in my comments here: https://github.com/thomaspoignant/go-feature-flag/pull/2734#issuecomment-2508689545

gogbog:
@thomaspoignant , you're right, I completely forgot about the requirement for having two flags with the same name. Unfortunately, if we want to support this, we would be forced to use multiple retrievers for each file, which leads to the challenge with dynamic receivers and redeployment. Currently, we're using MongoDB as a retriever, and I'm curious how would the JSON representation look like if we choose to implement **Solution 1** and would it be possible?

thomaspoignant:
>Currently, we're using MongoDB as a retriever, and I'm curious how would the JSON representation look like if we choose to implement Solution 1 and would it be possible?

Yes you're right this solution since not the best one with MongoDB and other DBs.
The more I look at it, the more solution 2 seems to be the way to go.

tomflenner:
Hey there !

I am bringing our use case here to see if we can find an elegant solution with this flagSet feature to answer the problem !

We're using GOFF from relayproxy with a HTTP Retriever that is a simple endpoint which build a JSON payload based on stuff in database. The feature flag data are modelized in database by feature and group up by stuff like userId, clientId etc.

The problem we're facing is that we are multitenant and we might dont want mix up all our client feature flag in a single configuration when the relayproxy fetch data from the HTTP Retriever.

Do you have more details on how this flagSet will be used with HTTP Retriever ? Do we have a way to specify for example per instance or per client the set of data that is needed to being retrieved from the HTTP Retriever ?

thomaspoignant:
>Do you have more details on how this flagSet will be used with HTTP Retriever ? Do we have a way to specify for example per instance or per client the set of data that is needed to being retrieved from the HTTP Retriever ?

Yes your point is similar to the one of @iisharankov and I get it completely.
I want to proceed in 2 steps for the flag sets
1. Enable flag sets in GO Feature Flag _(in a static way)_.
2. Be able to dynamically generate flag sets from a single provider.

thomaspoignant:
Based on your feedback, I have proposed a [new solution](https://github.com/thomaspoignant/go-feature-flag/blob/spec-flag-set/website/src/pages/specification/20241027-flagsets.md#solution-3) that seems more aligned with the concerns that some of you have raised.

I'll be happy to get your feedback on this one.
cc @tomflenner @gogbog @luizgribeiro @iisharankov

thomaspoignant:
I also have an extra question, how do you think the `exporter` and `notifier` should work with a flag set?
- Should we export everything to the same destination? 
- Should we notify all the flag changes or `notify` only per flag set?

thomaspoignant:
Another idea of configuration that came to my mind (and that is probably the best one) is something like this:

Where `type: static` means that the retriever will return only 1 flagset and `type: dynamic` can return multiple flagset.
```yaml
listen: 1031
pollingInterval: 10000
startWithRetrieverError: false

# default flag set (no need to specify anything)
retrievers:
  - kind: file
    path: /xxx/flags.goff.json
exporter:
  kind: log
notifier:
  - kind: discord
    webhookUrl: https://xxx.xxx
  - kind: microsoftteams
    webhookUrl: https://xxx.xxx


# static flag set 
flagSets:
  - type: static
    name: teamA
    retrievers:
      - kind: file
        path: /xxx/flags2.goff.json 
    notifier:
      - kind: discord
        webhookUrl: https://xxx.xxx
      - kind: microsoftteams
        webhookUrl: https://xxx.xxx
    exporter:
      kind: log
  
  - type: dynamic
    retrievers:
      - kind: file
        path: /xxx/flags3.goff.json
    notifier:
      - kind: discord
        webhookUrl: https://xxx.xxx
      - kind: microsoftteams
        webhookUrl: https://xxx.xxx
    exporter:
      kind: log

  - type: dynamic
    retrievers:
      - kind: file
        path: /xxx/flags4.goff.json
    notifier:
      - kind: discord
        webhookUrl: https://xxx.xxx
      - kind: microsoftteams
        webhookUrl: https://xxx.xxx
    exporter:
      kind: log
```

website/src/pages/specification/20241027-flagsets.md
```diff
@@ -0,0 +1,269 @@
...
+- It is easy to understand which flags are part of a flag set.
+- It is easy to give ownership of a flag set to a team, by giving them access to the file.
+- The flag set name is directly associated to the retriever, so it is not possible to edit it directly inside the file.
+
+**CON**
+
+
+---
+
+### Solution 3
```
luizgribeiro, line 181: This one sounds really good!

Even though it's more complex you can still have the best out of both approaches.

If the precedence between strategies is well documented I don't see much problem in this.

There could also be something like, depending on the `kind` of the retriever, the existence of a flag set configuration be required

gogbog:
@thomaspoignant, for us, the main advantage will be the dynamic flagsets since we store all flags in a single database. The only way I see the static solution working with our system is if we reinitialize the pod every time a new team is added and dynamically update the GOFF configuration (somehow). However, this would likely result in multiple connections to the database even if it works.

I was wondering are you still thinking about the idea and if yes, if we choose the dynamic route, how do you imagine the flag configuration to look like?

This is an example flag in a database:
```
{
  "flag": "new-admin-access",
  "variations": {
    "default_var": false,
    "false_var": false,
    "true_var": true
  },
  "defaultRule": {
    "percentage": {
      "false_var": 70,
      "true_var": 30
    }
  }
}
```

I noticed that you preferred keeping the name within the actual object rather than using the usual approach of having the flag name as the key and the flag configuration as the value. Would it be possible to include the flagset name inside the object as well, or were you considering a different approach?

Also I was also wondering - how will we link the authentication token to the flagset the user is authorized to access?

thomaspoignant:
I am closing this one since we have implemented a first version of the flag set capability #3610.
This is a beta version for now, but feel free to provide feedback or open an issue to provide feedback.

PR #2751, opened 2024-12-04
fix(exporter): Ensure that timestamp are in second and not milliseconds

thomaspoignant:
## Description
This PR is adding a conversion check on the feature event date to be sure that what we receive is in seconds not in milliseconds. 

## Checklist
- [x] I have tested this code


cmd/relayproxy/controller/collect_eval_data.go
```diff
@@ -60,6 +60,14 @@ func (h *collectEvalData) Handler(c echo.Context) error {
 		if event.Source == "" {
 			event.Source = "PROVIDER_CACHE"
 		}
+		// force the creation date to be a unix timestamp
+		if event.CreationDate > 9999999999 {
+			// if we receive a timestamp in milliseconds, we convert it to seconds
+			// but since it is totally possible to have a timestamp in seconds that is bigger than 9999999999
+			// we will accept timestamp up to 9999999999 (2286-11-20 18:46:39 +0100 CET)
+			event.CreationDate, _ = strconv.ParseInt(
+				strconv.FormatInt(event.CreationDate, 10)[:10], 10, 64)
```
augustinasrce, line 69: logging some warning would help resolve such issues in the clients. At least it would be visible.

thomaspoignant: Yes you are right, let me add a WARN log.

augustinasrce (approved):

PR #2756, opened 2024-12-06
docs(example): Fixing link in react example

### LINKED ISSUE #2755: (doc) react example seems outdated and not very clear
### Change in the documentation

in README.md links are not working 

https://github.com/thomaspoignant/go-feature-flag/blob/main/examples/openfeature_react/react-app/src/App.tsx

Implementation it self is not linted/formatted properly
Initialization of OpenFeature sdk is a bit confusing and could be better. 

maybe reusing this would be beneficial 
https://github.com/open-feature/js-sdk/blob/main/packages/react/README.md

thomaspoignant:
## Description
- Fixing link in react example
- Format code.

## Closes issue(s)
Resolve #2755

## Checklist
- [x] I have tested this code


examples/openfeature_react/react-app/src/App.tsx
```diff
@@ -15,35 +15,35 @@ OpenFeature.setProvider(goFeatureFlagWebProvider);
...
   const setRandomEvaluationContext = () => {
     const availableContexts: Array<EvaluationContext> = [
-      { targetingKey: "user-2", userType: "dev", email: "contact@gofeatureflag.org" },
-      { targetingKey: "user-3", userType: "admin", company: "GO Feature Flag" },
-      { targetingKey: "user-4", userType: "customer", location: "Paris" },
-      { targetingKey: "user-5" }
+      {targetingKey: "user-2", userType: "dev", email: "contact@gofeatureflag.org"},
+      {targetingKey: "user-3", userType: "admin", company: "GO Feature Flag"},
+      {targetingKey: "user-4", userType: "customer", location: "Paris"},
+      {targetingKey: "user-5"}
```
augustinasrce, line 33: It's a bit unrealistic example. You would never load more than one context. At least I can't think of a use case. 

I would really like to split App and Page, so init and usage would be clearly distinguished. I get the idea of "lets show everything in one place", but I would say it's a bit hard to understand what is what.

thomaspoignant: Ok I got it, I will rework the example to make it more explicit.

PR #2790, opened 2024-12-12
Feat/Add PostgreSQL Retriever

gogbog:
## Description

This PR adds a PostgreSQL retriever for Goff. For example, we use PostgreSQL in our projects, and it would be nice to stick to one database for everything. Most likely, there are other people who would benefit from this as well.

## Implementation Details:

New Libraries:
- I used PGX V5 for the database connection and connection pool.
- Integration tests were done using testcontainers-go to create a PostgreSQL container instance

Current Retriever Options:
- URI - Database connection string
- Table - The table from which the flags will be fetched.
- Column - The column from which the flags will be fetched.
- Type - Currently hardcoded to `JSON`. In the future, we can extend this to support a more structured table approach.

@thomaspoignant , happy to share this and open to improving it further if you have any feedback!

## Checklist
- [x] I have tested this code
- [x] I have added example in the `examples` folder of how to run the provider


retriever/postgresqlretriever/retriever.go
```diff
@@ -0,0 +1,119 @@
...
+
+		if val, ok := doc["flag"]; ok {
+			delete(doc, "flag")
+			if str, ok := val.(string); ok {
+				ffDocs[str] = doc
+			} else {
+				r.logger.Error("Flag key does not have a string value")
+			}
+		} else {
+			r.logger.Error("No 'flag' entry found")
```
thomaspoignant, line 108: **🤓 nitpick:** ‏I am not sure that it is really an error to have no flag.
```suggestion
			r.logger.Warn("No 'flag' entry found")
```

retriever/postgresqlretriever/retriever.go
```diff
@@ -0,0 +1,119 @@
...
+	status string
+	logger *fflog.FFLogger
+}
+
+func (r *Retriever) Init(ctx context.Context, logger *fflog.FFLogger) error {
+	r.logger = logger
+	if r.dbPool == nil {
+		r.status = retriever.RetrieverNotReady
+
+		poolConfig, err := pgxpool.ParseConfig(r.URI)
```
thomaspoignant, line 32: **❔ question:** ‏Do we need a connection pool here? We will have only 1 connection per retriever, I am not super familiar with `pgxpool` but what value is it bringing here?

retriever/postgresqlretriever/retriever.go
```diff
@@ -0,0 +1,119 @@
...
+		if err != nil {
+			// This will not be a connection error, but a DSN parse error or
+			// another initialization error.
+			r.status = retriever.RetrieverError
+			return err
+		}
+
+		pool, err := pgxpool.NewWithConfig(ctx, poolConfig)
```
thomaspoignant, line 40: **❔ question:** ‏Same question for `pgxpool`.

retriever/postgresqlretriever/retriever.go
```diff
@@ -0,0 +1,119 @@
...
+// If a document does not comply with the specification it will be ignored
+func (r *Retriever) Retrieve(ctx context.Context) ([]byte, error) {
+	query := fmt.Sprintf("SELECT %s FROM %s WHERE %s IS NOT NULL", r.Column, r.Table, r.Column)
+	rows, err := r.dbPool.Query(ctx, query)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	ffDocs := make(map[string]interface{})
```
thomaspoignant, line 83: **🎯 suggestion:** ‏it is not super explicit to me what `ffDocs` contains. I can't come with a better name, but it does not sound self explain what do we store in this `map`.

thomaspoignant (commented):
Hey @gogbog this is a super nice start, I have left some small nit comments.

As we said in slack I would also like to have the structured version too in the same PR if possible. WDYT?

retriever/postgresqlretriever/retriever.go
gogbog: Changed that.

retriever/postgresqlretriever/retriever.go
gogbog: You are right. Changed it to direct connection instead of a pool. If we have single connection it won't bring any value.

retriever/postgresqlretriever/retriever.go
gogbog: Yup, it's not the best variable name. I've updated it to `mappedFlagDocs` as it feels more self-explanatory. I'm still not sure what name would be the best, If you have better suggestions, let me know.

thomaspoignant:
Sorry for the delay to work on this PR, we had a lot of discussion with @gogbog about this pull request because of the SQL schema.
After a discussion we decided to go with a pragmatic version and to keep the database schema really simple.
I have open a new pull request with the implementation #3881 

The new provider is also flagset aware and can filter on different flagsets.

PR #3068, opened 2025-02-10
fix: OpenTelemetry trace propagation

### LINKED ISSUE #3042: (bug) Tracing: Incoming traceparent header is ignored
### Observed behaviour

When requesting an endpoint of the relay proxy with a `traceparent` header set according to the W3C tracecontext specification (the de facto standard for this), it ignores this incoming trace information and starts a new trace with a new ID. This leads to two traces that are not connected to each other and hence mostly worthless. 

### Expected Behavior

The span started by the HTTP server instrumentation of the relay proxy should have the same trace ID as specified in the header and set the parent span ID accordingly. This would ensure traces are complete and fully useful.

I'm no expert in Go, but it looks like the OpenTelemetry SDK for Go might not set the default propagators (`TextMapPropagator`) like other OpenTelemetry SDKs where it defaults to the one suited for `traceparent`. If no or only a no-op propagator is set this will lead to the observed behaviour, in which case the relay proxy should actively set at least one propagator to enable trace continuation.

### Steps to reproduce

1. Set up the relay proxy with tracing enabled, exporting the spans to a backend like Jaeger or HoneyComb.
2. Send a request with `traceparent` header set like this: `traceparent: 00-777788de6930ea4a4f1ee18c6a11f8d7-ca872024dec6eab6-01`
3. Check the trace ID from the header (`777788de6930ea4a4f1ee18c6a11f8d7`) and the backend, it won't show up under this ID, but a different one. Also, its root span will not specify any parent.

thomaspoignant:
## Description
As reported by @ybasket, there is an issue in propagating the traces across different services.

This PR configure the propagation for Otel in order to follow the traces across multiple span.

Screenshot 2025-02-10 at 18 50 55

## Closes issue(s)
Resolve #3042

## Checklist
- [x] I have tested this code

### DIFF
--- a/cmd/relayproxy/api/opentelemetry/otel.go
+++ b/cmd/relayproxy/api/opentelemetry/otel.go
@@ -11,6 +11,7 @@ import (
 	"go.opentelemetry.io/contrib/samplers/jaegerremote"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/propagation"
 	"go.opentelemetry.io/otel/sdk/resource"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 	semconv "go.opentelemetry.io/otel/semconv/v1.26.0"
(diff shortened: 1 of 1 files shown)


hairyhenderson:
Ah! Thanks for this fix - I was just about to work on this after realizing it was broken 🤦‍♂️

PR #3141, opened 2025-02-24
feat(relayproxy): allow disable VersionHeader middleware

### LINKED ISSUE #3140: (feature) Disable version header in relay proxy
### Requirements

Hello !

We're using RelayProxy in production since several month now and our OPS teams reported that the `x-gofeatureflag-version` header was forward in HTTP requests. It would be nice to not have to tweak our Kong gateway to remove this header and let the component himself do it. Our team don't want to expose infrastructure concern inside the HTTP request for some reason.

After discussing with @thomaspoignant, this header is used to detect GOFF version and disable some feature in the provider for compatibility reason. At the moment he never needed to use this feature.

I suggest to allow users to deactivate this header from configuration to be more flexible on relay proxy usage.

tomflenner:
add a boolean EnableVersionHeader in configuration to conditionnally register the VersionHeader custom middleware and disable the header x-gofeatureflag-version

issue: https://github.com/thomaspoignant/go-feature-flag/issues/3140

## Description

Add a new boolean named `EnableVersionHeader` in documentation with default value set to `true` in order to not change initial behavior.

The idea behind this is to allow end user to disable `VersionHeader` custom middleware to remove `x-gofeatureflag-version` header.

## Closes issue(s)

Resolve #3140

## Checklist
- [X] I have tested this code
- [X] I have updated the documentation (`README.md` and `/website/docs`)

### DIFF
--- a/cmd/relayproxy/api/routes_monitoring.go
+++ b/cmd/relayproxy/api/routes_monitoring.go
@@ -18,9 +18,9 @@ func (s *Server) addMonitoringRoutes() {
 		s.monitoringEcho.Debug = s.config.IsDebugEnabled()
 		s.monitoringEcho.Use(custommiddleware.ZapLogger(s.zapLog, s.config))
 		s.monitoringEcho.Use(middleware.CORSWithConfig(middleware.DefaultCORSConfig))
-		if s.config.EnableVersionHeader {
-			s.monitoringEcho.Use(custommiddleware.VersionHeader(s.config))
-		}
+		s.apiEcho.Use(custommiddleware.VersionHeader(custommiddleware.VersionHeaderConfig{
+			RelayProxyConfig: s.config,
+		}))
 		s.monitoringEcho.Use(middleware.Recover())
 		s.initMonitoringEndpoint(s.monitoringEcho)
 	} else {
--- a/cmd/relayproxy/api/routes_monitoring.go
+++ b/cmd/relayproxy/api/routes_monitoring.go
@@ -18,7 +18,9 @@ func (s *Server) addMonitoringRoutes() {
 		s.monitoringEcho.Debug = s.config.IsDebugEnabled()
 		s.monitoringEcho.Use(custommiddleware.ZapLogger(s.zapLog, s.config))
 		s.monitoringEcho.Use(middleware.CORSWithConfig(middleware.DefaultCORSConfig))
-		s.monitoringEcho.Use(custommiddleware.VersionHeader(s.config))
+		if s.config.EnableVersionHeader {
+			s.monitoringEcho.Use(custommiddleware.VersionHeader(s.config))
+		}
 		s.monitoringEcho.Use(middleware.Recover())
 		s.initMonitoringEndpoint(s.monitoringEcho)
 	} else {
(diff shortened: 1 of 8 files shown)


tomflenner:
@thomaspoignant weird for codecov.

I added two test on `server_test.go` to check that server register and do not register the middleware based on 
`EnableVersionHeader` configuration but it do not recognize it a part of tested line.

Any idea on how to solve this ?

cmd/relayproxy/api/routes_monitoring.go
```diff
@@ -18,7 +18,9 @@ func (s *Server) addMonitoringRoutes() {
 		s.monitoringEcho.Debug = s.config.IsDebugEnabled()
 		s.monitoringEcho.Use(custommiddleware.ZapLogger(s.zapLog, s.config))
 		s.monitoringEcho.Use(middleware.CORSWithConfig(middleware.DefaultCORSConfig))
-		s.monitoringEcho.Use(custommiddleware.VersionHeader(s.config))
+		if s.config.EnableVersionHeader {
+			s.monitoringEcho.Use(custommiddleware.VersionHeader(s.config))
+		}
```
thomaspoignant, line 23: **🤓 nitpick:** ‏Instead of having a if, I would recommend to have a `Skipper` method as a constructor option to be able to follow the format of labstack/echo middlewares.

Having a struct as config object for the middleware, something like:

```go
type VersionHeaderConfig struct {
  // Skipper defines a function to skip middleware.
  Skipper Skipper
  // Relay proxy config
  RelayProxyConfig config.Config
}
```

cmd/relayproxy/config/config.go
```diff
@@ -262,6 +263,10 @@ type Config struct {
 	// Version is the version of the relay-proxy
 	Version string `mapstructure:"version" koanf:"version"`
 
+	// Enable x-gofeatureflag-version header in the relay-proxy HTTP response
+	// Default: true
+	EnableVersionHeader bool `mapstructure:"enableVersionHeader" koanf:"enableversionheader"`
```
thomaspoignant, line 268: **🤓 nitpick:** ‏Since the default behavior is to have the header enabled, I'll suggest to name this new option `DisableVersionHeader`.

thomaspoignant (commented):
Hey, thanks for your PR.

Considering that the default behavior is to have the middleware activated I would suggest to name the config property `DisableVersionHeader` instead and also I would try to stick to the common way practice of having `Skipper` in the middleware instead of a if around, which can allow to activate the middleware on certain patterns.

Also we will nee to update the documentation to mention this new option in this [documentation file](https://github.com/thomaspoignant/go-feature-flag/blob/main/website/docs/relay-proxy/configure-relay-proxy.mdx).

cmd/relayproxy/api/routes_monitoring.go
tomflenner: Done, thanks again for answering on slack about Skipper usage.

Did not know before.

thomaspoignant (approved):
@tomflenner Thanks a lot for this addition 🙏 

I have updated slightly the PR to be more idiomatic.
I will merge the PR as soon as the CI pass.

PR #3147, opened 2025-02-26
feat: handle 400 error with response message

gogbog:
## Description

The pull request improves error handling for the Python provider.

Currently, the error message is hardcoded. With this update, if a message parameter is sent in the response, it will be displayed."

Let me know if you'd like further refinements!


thomaspoignant (changes requested):
@gogbog Thanks for this pull request, this is really great.

**💭 thought:** ‏I think we should add a unit test to validate that it works as expected.
If you don't have time to work on it, let me know and I will do it.

openfeature/providers/python-provider/gofeatureflag_python_provider/provider.py
```diff
@@ -210,16 +211,29 @@ def generic_go_feature_flag_resolver(
...
                 if response.status == HTTPStatus.NOT_FOUND.value:
                     raise FlagNotFoundError(
                         "flag {} was not found in your configuration".format(flag_key)
                     )
 
-                if int(response.status) >= HTTPStatus.BAD_REQUEST.value:
+                # Handle 400 error code
+                if int(response.status) == HTTPStatus.BAD_REQUEST.value:
+                    response_dict = json.loads(response_body)
+                    errorMessage = response_dict.get("message")
```
thomaspoignant, line 225: **🤓 nitpick:** ‏to follow the rest of the naming convention it will be better to name the variable `error_message`.

gogbog:
> @gogbog Thanks for this pull request, this is really great.
> 
> **💭 thought:** ‏I think we should add a unit test to validate that it works as expected. If you don't have time to work on it, let me know and I will do it.

Thank you for your quick response. I considered writing tests, but I'm not sure at the moment how to trigger a 400 error from the GOFF side. The `OFREP` endpoints have more mature error handling, but in `v1`, it seems that errors only occur if the evaluation context is empty or if the key is not found in the body - both of which are already handled by the provider even before sending the request.

The main purpose of this PR is to make the provider more future-proof (e.g., for a potential migration to `OFREP`). Additionally, we have a wrapper for GOFF that adds extra logic and validations, returning a 400 error with a message. However, this message is not properly handled in the Python provider.

Let me know if you have any ideas on how we can test this, as I'm not completely sure at the moment.

thomaspoignant:
Ok got it, that's true that this is not easy to handle this.
I will merge it as is 👌

thomaspoignant (approved):

PR #3336, opened 2025-04-16
fix: a potential production situation

orenf:
## Description
When we deployed a service called debug_test_service, EKS generated an env var called DEBUG_TEST_SERVICE which
goff understood to mean enable debug, but the value of the service could not be coerced into boolean
so the goff service started crash looping.  This potentially exposed an additional security concern because any env var with this DEBUG prefix would be written to logs. This failure caused us a 2 hour production headache.

## Checklist
- [ ] I have tested this code

### DIFF
--- a/cmd/relayproxy/config/config.go
+++ b/cmd/relayproxy/config/config.go
@@ -162,7 +162,11 @@ func mapEnvVariablesProvider(log *zap.Logger) koanf.Provider {
 			return s, v
 		}
 
-		return strings.ReplaceAll(strings.ToLower(s), "_", "."), v
+		if strings.HasPrefix(s, "OTEL_") {
+			return strings.ReplaceAll(strings.ToLower(s), "_", "."), v
+		}
+
+		return strings.ToLower(s), v
 	})
 }
 


thomaspoignant (commented):
Hey thanks for your pull request and I am sorry that you had a production issue.
Unfortunately, this `ReplaceAll` is not linked to OTEL only but can be for other usecases.

I guess in your specific case it has enabled the debug, which is not supposed to happen.
The `debug` key is deprecated so it can be removed, but I think we need to think globally on how to avoid your situation.

I will try to work on this shortly for the next version.

orenf:
Thank you for clarifying @thomaspoignant.  While I haven't verified this, it seems perfectly conceivable that within a K8s ecosystem there is a potential for many env variable collisions with the **config**. For illustration of my point, consider a service called ENABLESWAGGER_123_SERVICE or LOGLEVEL_ABC.  If I'm understanding the functionality through the experience of this **debug** failure, then these collisions would result in similar failures. Please correct me if my understanding is incorrect. Thanks again.

thomaspoignant:
I am working on a fix.
What I plan to do is 
1. Remove the `debug` option (since this is deprecated and it was already plan to remove it see: https://github.com/thomaspoignant/go-feature-flag/blob/db17ea7f9793d9c364ea46ec30bf58d25af1aba6/cmd/relayproxy/config/config.go#L120-L126)

2. The second thing I will do, is allow to configure in the config file a prefix for the environment variable. Meaning that you can decide to use `YOURPREFIX_ENABLESWAGGER` instead of `ENABLESWAGGER`.

We won't change the default behavior since this is already used by a lot of customers in production already, but we will allow to add a prefix if necessary.

thomaspoignant:
@orenf I have open a new PR #3337 to solve your issue.

orenf:
I did not expect such a quick turnaround @thomaspoignant .  The solution looks good to me. Thank you again.

thomaspoignant:
@orenf It will be released in the next version that should come pretty soon.

thomaspoignant:
@orenf The fix is available in the `v1.44.0` that is currently being released.

PR #3413, opened 2025-05-09
feat: First step in-process evaluation

thomaspoignant:
## Description
GO Feature Flag for now works only with remote calls to the relay-proxy, all the evaluations are happening in the server not on the OpenFeature providers.

This PR is the first step to move server evaluation directly in the providers ensuring no issue with network calls and a consistent way between languages.

In this PR you will find 2 things:

### 1. `/v1/flag/configuration` endpoint
This endpoint is exposing your flag configuration, allowing the providers to retrieve all the configuration of your flags.

### 2. Webassembly build of the evaluation logic
A Webassembly component that we will be able to embed directly in the providers that will ensure we are doing the evaluation the same way in all the providers.

> [!WARNING]  
> This is only the first step, to have in-process evaluation, we will have to rewrite all the providers and embed the webassembly code there.
> If you want to look at how a provider with webassembly should work you can check this branch for the java provider https://github.com/open-feature/java-sdk-contrib/tree/goff-inprocess-eval/providers/go-feature-flag.

## Checklist
- [ ] I have tested this code


hoerup:
I assume this will present all defined flags to consumers - would it make sense to combine it with the "feature set" concept so clients won't need to load everything?

thomaspoignant:
@hoerup yes this is actually a good point and this is the goal.
I just want to avoid to link the development together too much, but flag sets will come for sure, I am just wondering how to do it well (specially in the case of dynamic flag sets).

PR #3656, opened 2025-07-16
fix(python): Handle missing anonymous attribute in Python provider data collector hook

### LINKED ISSUE #3640: (question) why is "anonymous" required?
### Your question

Hi,
It seems from the implementation of the python provider's data collector and error hook that the attribute "anonymous" is required. All the examples use it. However sometimes it might not be provided, in which case an error is thrown.
Reference:
https://github.com/thomaspoignant/go-feature-flag/blob/5fa8644001825c03afcc28e7c7a455ece40263f9/openfeature/providers/python-provider/gofeatureflag_python_provider/data_collector_hook.py#L76

As far as I understand this attribute is not specified by Openfeature. However it is specified by some vendors, such as [Launchdarkly](https://launchdarkly-python-sdk.readthedocs.io/en/latest/api-main.html#ldclient.ContextBuilder.anonymous). Not sure if it is required though with no default.

Is there perhaps a bug in the python provider that makes it a required attribute? Or is it intended?

thomaspoignant:
## Summary
- Fixed KeyError when 'anonymous' attribute is missing from evaluation context in Python provider data collector hook
- Added safe dictionary access using .get() method with default value of False
- Added comprehensive tests to verify the fix works correctly

## Changes
- Modified `data_collector_hook.py` to use `attributes.get("anonymous", False)` instead of `attributes["anonymous"]`
- Added two new tests to verify both `after` and `error` hook methods handle missing anonymous attribute correctly

## Test plan
- [x] All existing tests pass (37/37)
- [x] New tests verify the fix handles missing anonymous attribute correctly
- [x] When anonymous attribute is missing, user is treated as non-anonymous (contextKind="user")
- [x] When anonymous attribute is present, behavior remains unchanged

Fixes #3640

### DIFF
--- a/.gemini/config.yaml
+++ b/.gemini/config.yaml
@@ -1,5 +1,12 @@
-pull_request_opened:
-  # we disable the summary and code review for the pull request opened event
-  # because I want to be sure to trigger it on demand and not automatically.
-  summary: false
-  code_review: false
+have_fun: false
+code_review:
+  disable: false
+  comment_severity_threshold: MEDIUM
+  max_review_comments: -1
+  pull_request_opened:
+    # we disable the summary and code review for the pull request opened event
+    # because we want to be sure to trigger it on demand and not automatically.
+    help: false
+    summary: false
+    code_review: false
+ignore_patterns: []


kouk (commented):
LGTM!

PR #3910, opened 2025-09-16
feat: add evaluation rules name to all the flags evaluations

carlossilveirajunior:
## Description

Currently, on the single flags evaluation, it is possible to know what the rule was that generated the variant. That is not true for the all flags evaluation. This Pr changes data, making the rule name (if available) part of all flags metadata. 

## Closes issue(s)
N/A

## Checklist
- [X] I have tested this code
- [N/A] I have updated the documentation (`README.md` and `/website/docs`)


thomaspoignant (approved):
🚀 This PR looks great to me.
This is way better to have the computation of the metadata as close as the evaluation possible.

I've run the gemini code review, the points mentioned are really nit picking, but feel free to address them if you want.

I will release the PR in the next version.

PR #3911, opened 2025-09-16
feat: reuse retriever config logic from cmd/relayproxy

### LINKED ISSUE #3876: (feature) Allow cli tool to use retriever other than fileretriever
### Requirements

Currently `pkg/cli` tool is capable only to use local file while performing evaluation of flag(s). This is [hardcoded](https://github.com/thomaspoignant/go-feature-flag/blob/main/cmd/cli/evaluate/evaluate.go#L27) as follows

```go
func (e evaluate) Evaluate() (map[string]model.RawVarResult, error) {
	c := ffclient.Config{
		PollingInterval:       10 * time.Minute,
		DisableNotifierOnInit: true,
		Context:               context.Background(),
		Retriever:             &fileretriever.Retriever{Path: e.config},
		FileFormat:            e.fileFormat,
	}
```

I propose to expand the capability of this tool to support at least `httpretriever`. Ideally I think we should support [each](https://github.com/thomaspoignant/go-feature-flag/tree/main/retriever) retriever.

DeamonDev:
## Description
In order to refactor `cli` tool as mentioned in #3876 I plan to re-use `RetreiverConf` [struct](https://github.com/thomaspoignant/go-feature-flag/blob/main/cmd/relayproxy/config/retriever.go#L10) and initialize proper retriever in `cli`. I think it is good idea since thanks to that we'll avoid code duplication and we'll reuse `IsValid` [method](https://github.com/thomaspoignant/go-feature-flag/blob/main/cmd/relayproxy/config/retriever.go#L42) defined on `*RetrieverConf`. 

## Closes issue(s)
This is the half-step in order to close #3876. The second one will be direct refactor in `cli` package.

## Checklist
- [x] I have tested this code
- [N/A] I have updated the documentation (`README.md` and `/website/docs`)


thomaspoignant:
Hey thank for this pull request and this is great to reuse the config logic.

Despite that, I am not sure if moving the config to the retriever is the best place since it is used only by the tools in the `cmd` package.
Instead of moving them into `retriever/` I would prefer if we can move them to a new directory in the root of the repo called 
`cmdhelpers/retrieverconf`. 

It will make clearer the purpose of the package because this is used only for the tools not for the GO module.

DeamonDev:
@thomaspoignant okay good point, I moved these into new directory

cmdhelpers/retrieverconf/init/retriever_init_internal.go
```diff
@@ -1,10 +1,8 @@
-package service
+package init
```
thomaspoignant: **🎯 suggestion:** ‏
I am not a big fan of splitting the logic of init retriever in 2 files.
Can you merge `retriever_init.go` and `retriever_init_internal.go`, it will be more consistent with the rest of the code base.

cmdhelpers/retrieverconf/init/retriever_init_internal.go
```diff
@@ -23,26 +21,27 @@ import (
 )
 
 // retrieverFactory defines the signature for retriever factory functions
-type retrieverFactory func(*config.RetrieverConf, time.Duration) (retriever.Retriever, error)
+type retrieverFactory func(*retrieverconf.RetrieverConf, *DefaultRetrieverConfig) (retriever.Retriever, error)
```
thomaspoignant: **🪄 polish:**
I see here that you know passing `defaultRetrieverConfig DefaultRetrieverConfig` instead of the timeout.

I don't like this approach, because in the end you are modifying multiple time the value of the struct containing the default values. _(See [here](https://github.com/thomaspoignant/go-feature-flag/pull/3911/files#diff-42f7c5b288ea10b7421a94c30b69b002ead62609ba0f70133bbc0706dcf572c3R21))_

I think the previous approach with passing the `timeout` was more readable and less prone to side effects.

```suggestion
type retrieverFactory func(*retrieverconf.RetrieverConf, time.Duration) (retriever.Retriever, error)
```

cmdhelpers/retrieverconf/init/retriever_init_internal.go
```diff
@@ -51,73 +50,78 @@ func createGitHubRetriever(c *config.RetrieverConf, timeout time.Duration) (retr
 		RepositorySlug: c.RepositorySlug,
 		Branch: func() string {
 			if c.Branch == "" {
-				return config.DefaultRetriever.GitBranch
+				return defaultRetrieverConfig.GitBranch
```
thomaspoignant: **🎯 suggestion:** ‏
Considering the comments [here](https://github.com/thomaspoignant/go-feature-flag/pull/3911/files#r2359718268) you can still use the previous approach use the public struct value directly here.


```suggestion
				return config.DefaultRetrieverConfig.GitBranch
```

cmdhelpers/retrieverconf/init/retriever_init.go
```diff
@@ -0,0 +1,28 @@
...
+
+type DefaultRetrieverConfig struct {
+	Timeout    time.Duration
+	HTTPMethod string
+	GitBranch  string
+}
+
+// InitRetriever initialize the retriever based on the configuration
+func InitRetriever(
+	c *retrieverconf.RetrieverConf, defaultRetrieverConfig DefaultRetrieverConfig) (retriever.Retriever, error) {
```
thomaspoignant: 🪄 polish:
Avoid passing the `defaultRetrieverConfig DefaultRetrieverConfig` param because we don't want it to be modified.

See https://github.com/thomaspoignant/go-feature-flag/pull/3911/files#r2359718268 for the context.

cmdhelpers/retrieverconf/init/retriever_init_internal.go
DeamonDev: Okay, I wondered about it. But I think as I dont pass the pointer to the method then I just modify the field of the local copy? Nevertheless I think you are right from the POV of code clarity.

thomaspoignant: Yes I think it will work fine but it will be easier to read.

DeamonDev:
@thomaspoignant I commited changes, thanks for the comments and links for the broader context, that was insightful for me.

thomaspoignant (approved):
We are good to go, I will merge the change.
Thanks a lot 🙇

PR #3986, opened 2025-09-30
feat: add k8s secret retriever

ndrshrzg:
## Description

Retrieving feature flags from Kubernetes ConfigMap is already supported. Support for [Kubernetes Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) is missing.

This PR adds a `k8sretriever.SecretRetriever` implementation of the `Retriever` interface. Also adds corresponding unit test.

## Checklist
- [ x ] I have tested this code
- [ x ] I have added unit test to cover this code
- [ x ] I have followed the [contributing guide](CONTRIBUTING.md)

### DIFF
--- a/retriever/k8sretriever/secret_retriever.go
+++ b/retriever/k8sretriever/secret_retriever.go
@@ -1,0 +1,59 @@
+package k8sretriever
+
+import (
+	"context"
+	"encoding/base64"
+	"fmt"
+
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


thomaspoignant:
Hey thanks for opening this pull request.
I have a question, in which case do you think using a secret is better than using configmaps? I can't see the value of storing feature flags inside secrets.

Do you have any use case where you thing this is better to have secrets instead of confimaps?

thomaspoignant:
Due to the lack on answer I will close this pull request.

Issue #4024, opened 2025-10-06
(change) Initialization of retrievers should also eventually call their specific initialization methods


DeamonDev:
### Motivation

I observed the flow of initializing of some retrievers may be not complete. When we call [InitRetriever](https://github.com/thomaspoignant/go-feature-flag/blob/main/cmdhelpers/retrieverconf/init/retriever_init.go#L46) method to spawn 
`s3` retriever then we internally call 

```go
func createS3Retriever(c *retrieverconf.RetrieverConf, _ time.Duration) (retriever.Retriever, error) {
	awsConfig, err := awsConf.LoadDefaultConfig(context.Background())
	return &s3retrieverv2.Retriever{Bucket: c.Bucket, Item: c.Item, AwsConfig: &awsConfig}, err
}
```

But there is also [initialization method](https://github.com/thomaspoignant/go-feature-flag/blob/main/retriever/s3retrieverv2/retriever.go#L40) for `s3` retriever.

### Requirements

Discussion whether we should call specific init methods here or not

thomaspoignant:
@DeamonDev GOFF is actually supporting multiple interfaces as you see [here](https://github.com/thomaspoignant/go-feature-flag/blob/main/retriever/retriever.go).

Depending on which interface the retriever is using it will call the Init method of the retriever _(see [here](https://github.com/thomaspoignant/go-feature-flag/blob/aed9fa017d474f4e6f8117bf58d9184c197be137/retriever/manager.go#L95))_.

The code snippet you've put in the issue is used only in the relay proxy, that force to use the AWS default locations for the authentication. 
For the `cli` we should probably have the same mechanism because it is more secure than providing the secrets to the cli directly.

DeamonDev:
Thanks for pointing this out. I added this additional init methods analogously to goff in #4025.

thomaspoignant:
Closing this issue since there is not really an issue here.

Issue #4023, opened 2025-10-06
(change) Add serializable configuration layer for redis options in struct describing configuration of retriever


DeamonDev:
### Motivation

Currently, the struct `RetrieverConf` contains the field `RedisOptions *redis.Option`, which is not serializable into JSON, since it contains function fields. Maybe we should create some kind of a "serializable layer" ? In that case, it would be easier pass command line configuration properties in `cmd/cli` package (see: #4025 for ongoing implementation)

### Requirements

Discussion on that matter

thomaspoignant:
@DeamonDev thanks for this issue, I think you are right we should probably have a serializable layer, my only worries is having to support a layer on GOFF side.

But that being said I don't see any other way to do it. So having something simple like bellow would be ok.
Do you think we need more than that?

```go
type RedisConfig struct {
    Addr     string `json:"addr"`
    Username string `json:"username"`
    Password string `json:"password"`
    DB       int    `json:"db"`
}
```

DeamonDev:
Yeah, I think we agree on that. I think its better to do this work incrementally and **lay back** refactor of this configuration struct in the ongoing cli refactor #4025. Then after refactoring this struct, we could add new PR with lacking redis support.

PR #4025, opened 2025-10-06
Feature add support for all retrievers in cli

### LINKED ISSUE #3876: (feature) Allow cli tool to use retriever other than fileretriever

DeamonDev:
## Description

I added new flags related to other kind of retrievers. Based on them the new retriever is being spawned (but see #4024). 
There is also new `--check-mode` flag which when present would only dump the configuration of retriever being used to perform `evaluate` call. It also served me while testing some multi-flags (like http's `--header`) because there is some kind of parsing logic at `evaluate_cmd` layer.

The `redis` retriever is not supported because of #4023. We may of course hack it and provide empty output when `--kind=redis` but I think the refactor of `RetrieverConf` struct would benefit more. 

## Closes issue(s)

Resolve #3876 

## Checklist
- [ ] I have tested this code


retriever/s3retrieverv2/retriever.go
```diff
@@ -97,3 +97,7 @@ func (s *Retriever) Retrieve(ctx context.Context) ([]byte, error) {
 
 	return writerAt.Bytes(), nil
 }
+
+func (s *Retriever) SetDownloader(downloader DownloaderAPI) {
+	s.downloader = downloader
+}
```
thomaspoignant, line 103: Just to be sure, why do you need this function?
You can probably set the `downloader` directly in the struct?

DeamonDev: It is used indeed in test

```go
r, err := retrieverInit.InitRetriever(
	&retrieverconf.RetrieverConf{
		Kind:   "s3",
		Bucket: "Bucket",
		Item:   "valid",
	})
if err != nil {
	return evaluate{}, err
}
s3Retriever, _ := r.(*s3retrieverv2.Retriever)
s3Retriever.SetDownloader(downloader)
```

And `downloader` is not (and should not be) a member of `RetrieverConf` (here I used, of course mocked downloader)

cmd/cli/evaluate/evaluate_cmd.go
```diff
@@ -66,3 +244,49 @@
...
+		}
+		if len(parts) != 2 {
+			continue
+		}
+		key := strings.TrimSpace(parts[0])
+		val := strings.TrimSpace(parts[1])
+		result[key] = append(result[key], val)
+	}
+	return result
+}
```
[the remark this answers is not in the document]
DeamonDev: good catch

cmd/cli/README.md
```diff
@@ -4,40 +4,194 @@
...
+- `s3`
+- `googleStorage`
+- `configmap` (kubernetes)
+- `mongodb`
+- `bitbucket`
+- `azureBlobStorage`
+- `postgresql`
+
+**Caution**: We do not support `redis` retriever as for now due
+to: https://github.com/thomaspoignant/go-feature-flag/issues/4023.
```
thomaspoignant, line 69: Do you want to add the serializable layer on this PR or do you prefer to open another pull request for this?

cmd/cli/README.md
```diff
@@ -4,40 +4,194 @@
...
+depends on that one, for
+example `--path` parameter is used to specify local file when `--kind` being `file` but when `--kind` being `github` it
+is used to specify
+the path to the remote file.
+
+#### File
+
+| Flag     | Description                               | Default |
+|----------|-------------------------------------------|---------|
+| `--path` | Path to the local flag configuration file | ""      |
```
thomaspoignant, line 83: Can we have a default value to file to be similar to the old behavior as much as possible?

cmd/cli/README.md
```diff
@@ -4,40 +4,194 @@
...
+the following command:
+
 ```shell
 go-feature-flag-cli evaluate --config="<location_of_your_flag_configuration_file>" --flag="<name_of_your_flag_to_evaluate>" --ctx='<evaluation_ctx_as_json_string>'
 ```
 
+may be replaced by:
+
+```shell
+go-feature-flag-cli evaluate --kind="file" --path="<location_of_your_flag_configuration_file>" --flag="<name_of_your_flag_to_evaluate>" --ctx='<evaluation_ctx_as_json_string>'
```
thomaspoignant, line 179: **👏 praise:** ‏This is great to have an example here.

cmdhelpers/retrieverconf/retriever_conf.go
```diff
@@ -20,46 +20,45 @@ var DefaultRetrieverConfig = struct {
...
+	HTTPBody    string              `mapstructure:"body"          koanf:"body" json:"HTTPBody"`
+	HTTPHeaders map[string][]string `mapstructure:"headers"       koanf:"headers" json:"HTTPHeaders"`
+	Bucket      string              `mapstructure:"bucket"        koanf:"bucket" json:"Bucket"`
+	Object      string              `mapstructure:"object"        koanf:"object" json:"Object"`
+	Item        string              `mapstructure:"item"          koanf:"item" json:"Item"`
+	Namespace   string              `mapstructure:"namespace"     koanf:"namespace" json:"Namespace"`
+	ConfigMap   string              `mapstructure:"configmap"     koanf:"configmap" json:"ConfigMap"`
+	Key         string              `mapstructure:"key"           koanf:"key" json:"Key"`
+	BaseURL     string              `mapstructure:"baseUrl"       koanf:"baseurl" json:"BaseURL"`
+	AuthToken   string              `mapstructure:"token"         koanf:"token" json:"AuthToken"`
```
thomaspoignant: Do you really need to add the `json` tag in the struct?
What is the reason do you need it?

cmd/cli/evaluate/evaluate_cmd.go
```diff
@@ -1,57 +1,264 @@
...
+				Container:   container,
+				AccountName: accountName,
+				AccountKey:  accountKey,
+				Table:       table,
+				Columns:     parsedColumns,
+			}
+
+			err := retrieverConf.IsValid()
+			if err != nil {
+				return err
```
thomaspoignant, line 146: **🔨 issue:** ‏ With the current cli flags names there is an issue when the cli is in error, the error message is not consistent.

For example if I run 
```shell
cli evaluate --kind=github --ctx='{"targetingKey": "123"}'
```

I will have the following error:
```
  ERROR   error executing command: invalid retriever: no "repositorySlug" property found for kind "github"
```

But your cli flags are waiting for `repository-slug` and not `repositorySlug`.

cmd/cli/README.md
DeamonDev: I mentioned in the issue I would like to separate these. WDYT? We may discuss it here ofc

cmd/cli/README.md
DeamonDev: I don't think I understand what u mean. I set default value of `kind` to be `file`, so we're backward compatible (there is a test for that)

cmd/cli/evaluate/evaluate_cmd.go
DeamonDev: Yeah, sth is wrong with that. I'll definitely look after it.

cmdhelpers/retrieverconf/retriever_conf.go
DeamonDev: I think I had some problem without them. I'll make sure about it.

cmd/cli/evaluate/evaluate_cmd.go
DeamonDev: Okay, these messages come from `validateXYZRetriever`'s methods defined on `*RetrieverConf` struct. For example:

```golang
func (c *RetrieverConf) validateGitRetriever() error {
	if c.RepositorySlug == "" {
		return fmt.Errorf(
			"invalid retriever: no \"repositorySlug\" property found for kind \"%s\"",
			c.Kind,
		)
	}
	if c.Path == "" {
		return fmt.Errorf("invalid retriever: no \"path\" property found for kind \"%s\"", c.Kind)
	}
	return nil
}
```

Hence, it is kind of a consequence of [re-using the logic](https://github.com/thomaspoignant/go-feature-flag/pull/3911) in `retriever_conf.go` in `cli` and `goff`.

cmdhelpers/retrieverconf/retriever_conf.go
DeamonDev: I checked it and removed these tags. I dont remember what kind of errors I had previously. But thanks for your alertness! :) 

edit: that was problem with linter `musttag`. I annotated now it with `// nolint:musttag`.

cmd/cli/README.md
thomaspoignant: This is all good no worries, we can address this later.

PR #4106, opened 2025-10-19
feat: Python Provider Graceful Exit

### LINKED ISSUE #4105: (change) Improving OpenFeature Python Provider: Better Exit Logic
### Motivation

The current Python provider has the following quirks:
- Requires manually calling `api.shutdown()` on exit, which means the user needs to handle "graceful exit" logic. This can be cumbersome especially when encountering an error. If the user did not implement "graceful exit" correctly, the program will hang in a "running" state due to cache invalidation and telemetry hook threads started in the goff provider doesn't stop automatically. This can be very misleading, as the OS/docker will think the program is still running when actually it's crashed or finished execution. 
- Even when the user correctly calls `api.shutdown()` on exit, the program will still hang for a while waiting for the telemetry hook thread to shut down. Ideally, we should immediately upload the cached telemetry and quit. 

### Requirements

- Register provider shutdown on program termination using `atexit`, even when the user doesn't manually run `api.shutdown()`. This will avoid program hangs even when an exception is raised. 
- Use [daemon threads](https://docs.python.org/3/library/threading.html#:~:text=A%20thread%20can%20be%20flagged%20as%20a%20%E2%80%9Cdaemon%20thread%E2%80%9D.%20The%20significance%20of%20this%20flag%20is%20that%20the%20entire%20Python%20program%20exits%20when%20only%20daemon%20threads%20are%20left.) to avoid running threads from blocking program termination ([`atexit` runs at python program termination](https://docs.python.org/3/library/atexit.html#:~:text=Functions%20thus%20registered%20are%20automatically%20executed%20upon%20normal%20interpreter%20termination), which means the daemon threads are still running, so it's safe to use `atexit` to clean up daemon threads). 
- Use `threading.Event` for `DataCollectorHook._thread_stopper` and `self._thread_stopper.wait(waiting_time)` instead of `time.sleep(waiting_time)` during polling. This will allow the wait to immediately return when the thread stopper is set, thus exiting immediately while keeping the behavior the same as `time.sleep` in other situations.

michaelmyc:
## Description

Ensures the Python provider have a graceful exit mechanism to close running threads at program exit. Also ensures polling threads are closed immediately rather than waiting for the next polling cycle. 

## Closes issue(s)

Resolve #4105

## Checklist
- [x] I have tested this code

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -111,6 +111,10 @@ To launch the tests please run:
 make test
 ```
 
+### AI review
+The repository is ready to receive gemini reviews.  
+If you want a fast feedback loop you can add a comment `/gemini review` in your pull request and gemini will review the pull request.
+
 ## 🤔 Where can I ask questions about the project?
 
 If you want to contribute and you have any questions you can use different ways to contact us.
--- a/examples/openfeature_react/react-app/package-lock.json
+++ b/examples/openfeature_react/react-app/package-lock.json
@@ -28,7 +28,7 @@
         "postcss": "^8.5.6",
         "tailwindcss": "^4.1.14",
         "typescript": "^5.9.3",
-        "vite": "^7.1.10"
+        "vite": "^7.1.11"
       }
     },
     "node_modules/@babel/code-frame": {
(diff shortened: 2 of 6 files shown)


michaelmyc:
Still need a few tests to make sure:
1. The graceful exit mechanism works without needing to call `api.shutdown()`
2. The graceful exit won't do any harm when `api.shutdown()` is called
3. The event trigger closes the hook's thread before wait time has expired

Will add those tests once I get some time on my hand.

thomaspoignant (approved):
@michaelmyc that looks good to me.
It will be part of the next release of the provider.

PR #4261, opened 2025-11-05
feat: Start the relayproxy as a unix socket.

### LINKED ISSUE #4244: (feature) Add Unix socket support for GO Feature Flag relay proxy
### Motivation

Currently, the GO Feature Flag relay proxy only supports HTTP/HTTPS connections over TCP. In Kubernetes environments, using Unix sockets can provide significant performance benefits by reducing network latency when the relay proxy is deployed as a sidecar container.

### Requirements

Expose the relay proxy API through Unix sockets in addition to the existing HTTP/HTTPS endpoints. This would enable:

1. **Reduced Network Latency**: Unix sockets provide faster inter-process communication compared to TCP/IP when processes are running on the same host
2. **Simplified Kubernetes Deployments**: When using the relay proxy as a sidecar, applications can communicate via Unix sockets without exposing additional network ports
3. **Improved Security**: Unix sockets provide better isolation as they operate at the filesystem level rather than network level

### Use Case

**Primary scenario**: Kubernetes deployments with sidecar pattern
- Deploy GO Feature Flag relay proxy as a sidecar container alongside the main application container
- Share a volume between containers to expose the Unix socket
- Application communicates with the relay proxy via Unix socket instead of localhost HTTP
- Eliminates network stack overhead for feature flag evaluations

### Proposed Implementation

The relay proxy should support a new configuration option to listen on a Unix socket in addition to (or instead of) TCP ports:

```yaml
# Example configuration
listen: 1031 #old option
unixSocket: "/var/run/goff/goff.sock"  # New option
```

When a Unix socket path is configured:
- The relay proxy should create and listen on the specified Unix socket
- All existing API endpoints should be accessible through the Unix socket
- The socket file should have appropriate permissions (configurable)
- Proper cleanup of the socket file on shutdown

### Technical Considerations

- Unix sockets are supported by Go's \`net\` package (via \`net.Listen("unix", path)\`)
- The Echo framework (used by the relay proxy) supports Unix socket listeners
- Socket file permissions and cleanup should be handled properly
- Documentation should include examples for Kubernetes deployment with shared volumes
- Consider supporting both TCP and Unix socket simultaneously for migration scenarios

### Expected Behavior

Clients should be able to make the same HTTP requests to the Unix socket as they would to the TCP endpoint:

```bash
# Current TCP approach
curl http://localhost:1031/health

# New Unix socket approach  
curl --unix-socket /var/run/goff/goff.sock http://localhost/health
```

---

**Context**: This feature was requested by @Augustinas on Slack to reduce network latency in Kubernetes environments.

thomaspoignant:
## Description
This PR is here to support starting the relay proxy as a unix socket.

The PR contains a new format for the configuration of the relayproxy.
Moving the `listen` option into a `server` option, allowing to have a more fine grained configuration.

 

## Closes issue(s)
Resolve #4244

## Checklist
- [x] I have tested this code

### DIFF
--- a/cmd/relayproxy/config/config_test.go
+++ b/cmd/relayproxy/config/config_test.go
@@ -524,7 +524,7 @@ func TestParseConfig_fileFromFolder(t *testing.T) {
 
 func TestConfig_IsValid(t *testing.T) {
 	type fields struct {
-		ListenPort              int
+		Server                  config.Server
 		HideBanner              bool
 		EnableSwagger           bool
 		Host                    string
@@ -552,13 +552,16 @@ func TestConfig_IsValid(t *testing.T) {
 		},
 		{
 			name:    "invalid port",
-			fields:  fields{ListenPort: 0},
+			fields:  fields{},
 			wantErr: assert.Error,
 		},
 		{
 			name: "no retriever",
 			fields: fields{
-				ListenPort: 8080,
+				Server: config.Server{
+					Port: 8080,
+					Mode: config.ServerModeHTTP,
+				},
 				Notifiers: []config.NotifierConf{
 					{
 						Kind:        "webhook",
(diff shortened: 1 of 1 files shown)


thomaspoignant:
_The CI failing for the Helm Check is normal because we need to release the new version of the relay proxy to make the changes works._

Issue #4067, opened 2025-10-13
(feature) Expose flag_name label on all_flag_evaluations_total


MattIPv4:
### Requirements

👋 As discussed in Slack, it would be useful to me if there were a metric exposed that tracked the bulk evaluations for each flag. While this would essentially expose the current flag config as a metric, I believe it may be useful to myself and others to be able to see at any given point in time, via the metrics, which flags are being sent to clients during bulk evaluations.

This will be rather noisy though, massively increasing the cardinality of `all_flag_evaluations_total`, so my suggestion as part of this would be to expose a new flag/setting that enables the additional `flag_name` label on `all_flag_evaluations_total`, so folks can choose to opt-in to this additional cardinality if they so desire, or to keep the current single-counter functionality of this metric.

thomaspoignant:
For the context, the slack discussion is [here](https://gophers.slack.com/archives/C029TH8KDFG/p1759954799548829).

PR #4350, opened 2025-11-22
feat(BulkEvaluate): add support for enabling metrics for each flag when using bulk evaluate

mooyg:
## Description
- Adds support for enabling per flag BulkEvaluation metrics
- Opt-in via `EnableBulkMetricFlagNames` config
- New metric name `all_flags_evaluations_total_with_flag`
- Refactors `NewMetric` to accept opts (backward-compatible)
## Closes issue(s)
Resolve [#4067](https://github.com/thomaspoignant/go-feature-flag/issues/4067)

## Checklist
- [x] I have tested this code


cmd/relayproxy/ofrep/evaluate.go
```diff
@@ -147,6 +147,7 @@ func (h *EvaluateCtrl) Evaluate(c echo.Context) error {
 // @Produce     json
 // @Accept	 	json
 // @Param       If-None-Match header string false "The request will be processed only if ETag doesn't match."
+// @Param       X-Enable-Bulk-Metric-Flag-Names header boolean false "If true, enables per flag metrics"
```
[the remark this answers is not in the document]
mooyg: For reviewer - I agree with this could you point me where I should place this if needed.

cmd/relayproxy/ofrep/evaluate_test.go
```diff
@@ -94,6 +95,20 @@ func Test_Bulk_Evaluation(t *testing.T) {
...
+				configFlagsLocation: configFlagsLocation,
+				headers: map[string]string{
+					"X-Enable-Bulk-Metric-Flag-Names": "true",
+				},
+			},
+			want: want{
+				httpCode: http.StatusOK,
+				bodyFile: "../testdata/ofrep/responses/valid_response.json",
+			},
+		},
```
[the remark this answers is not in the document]
mooyg: I’m open to doing this, but I don’t see other tests following that pattern.
A better approach would be to return the header in the response when it’s set to true, and assert that in the test. 

Let me know if we're cool with that

cmd/relayproxy/ofrep/evaluate.go
MattIPv4: I'm not sure a header is really the right way to do this, as this could allow clients to increase metric cardinality when the server operator doesn't want that? I would think it should be a CLI flag on the server, so the server operator controls the choice?

cmd/relayproxy/ofrep/evaluate.go
```diff
@@ -195,6 +197,9 @@ func (h *EvaluateCtrl) BulkEvaluate(c echo.Context) error {
 		allFlagsResp = flagset.AllFlagsState(evalCtx)
 	}
 	for key, val := range allFlagsResp.GetFlags() {
+		if enableBulkMetricFlagNames {
+			h.metrics.IncFlagEvaluation(key)
```
MattIPv4: Per #4067, the idea was to add a metric label name to the `all_flag_evaluations_total` metric, not just increment the regular eval metric, so that there are still two distinct metrics for bulk evals and regular evals.

mooyg: misunderstood, I think it makes sense to add a new metric `all_flag_evaluations_total_with_flag` in that case?

mooyg: I think this way this wouldn't even need to be opt-in. 

I'd prefer this but let me know your thoughts

MattIPv4: Yah I think a new metric name also works. I think it probably should still be opt-in (or at least opt-out) as it could be a very high cardinality metric for folks with large configs.

cmd/relayproxy/metric/metrics.go
```diff
@@ -6,12 +6,27 @@ import (
...
+	// default opts
+	mo := &MetricsOpts{
+		EnableBulkMetricFlagNames: false,
+	}
+
+	for _, opt := range opts {
+		if opt.EnableBulkMetricFlagNames {
+			mo.EnableBulkMetricFlagNames = true
+		}
+	}
```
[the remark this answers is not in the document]
mooyg: I think before is more clear - I have applied the suggestion regardless as of now.

thomaspoignant (approved):
It looks good to me, thanks a lot for your contribution 🙌.
_I have just added some documentation about this new option._

PR #4366, opened 2025-11-25
chore(wasm): Build for `wasm-unknown` instead of gojs

andreaTP:
## Description
This changes the build to wasm to be targeting wasm-unknown the most "platform agnostic" compilation target.

I attempted to quickly test the result in `open-feature/java-sdk-contrib` but one test seems to be looping, I haven't narrowed down the root cause yet.

## Closes issue(s)

Resolve #

## Checklist
- [ ] I have tested this code

### DIFF
--- a/cmd/wasm/main.go
+++ b/cmd/wasm/main.go
@@ -10,6 +10,8 @@ import (
 	"github.com/thomaspoignant/go-feature-flag/modules/core/model"
 	"github.com/thomaspoignant/go-feature-flag/modules/core/utils"
 	"github.com/thomaspoignant/go-feature-flag/modules/evaluation"
+
+	"unsafe"
 )
 
 // main is the entry point for the wasm module.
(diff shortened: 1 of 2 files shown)


cmd/wasm/main.go
```diff
@@ -73,3 +75,31 @@ func convertEvaluationCtx(ctx map[string]any) (ffcontext.Context, error) {
 	return ffcontext.NewEvaluationContextBuilder("").Build(),
 		fmt.Errorf("targetingKey not found in context")
 }
+
+
+// alloc/free implementation for wasm-unknown target from:
```
andreaTP, line 80: When not targeting wasm-wasi or gojs tinygo doesn't exports those functions.
Here we are exporting stable identifiers to be used on any target.

thomaspoignant: oh I didn't know about that, thanks for the addition.
I will test the full pipeline locally with your 2 PRs tonight or tomorrow.

andreaTP: Let me know how it goes and we continue from there :+1:

thomaspoignant:
@andreaTP I've done some tests and I end up with errors like `out of bounds memory access: attempted to access address: -1812927666 but limit is: 524288 and size: 1` not sure if this is the way I use chicory.

andreaTP:
@thomaspoignant which version of Java are you using? I suggest you try with Java21+ as a first iteration.

Also we might need to increase the default memory limits:

```java
Instance.builder().withMemoryLimits(new MemoryLimits(10))
```
"should" make the trick, if you have a reproducer I can take a look :-)

PR #4393, opened 2025-12-02
feat: shared pgx-pool

### LINKED ISSUE #4392: (change) Refactor PostgreSQL retriever to use shared pgxpool.Pool
### Motivation

The current PostgreSQL Retriever implementation utilizes `pgx.Conn` for database access. This is problematic because the system initialises a separate database connection for each flagset configured.

1. Resource Bottleneck: Every Retriever instance establishes and maintains its own dedicated `pgx.Conn`. If a user configures 100 or 200 flagsets, this leads to connection sprawl, potentially creating hundreds of independent database connections.

2. Connection Overhead: Each active database connection is resource-intensive. When managing large numbers of them, we end up using the database in a way it wasn’t designed for. This forces us to scale vertically and consume more and more resources.

### Requirements

The primary goal is to reduce the excessive number of connections and improve resource efficiency, we must migrate to a shared, global `pgxpool.Pool` and use it across all Postgres retrievers.

This change is supposed to do the following:

1. Single Pool: The entire application uses one high-performance connection pool, reusing a limited set of connections across all PostgreSQL Retriever instances, regardless of how many flagsets are configured.
2. Safe Cleanup: The pool only closes when the last active Retriever shuts down.

gogbog:
### Description

This PR addresses an issue with how database connections are managed when using multiple Postgres retrievers.
Currently, each retriever creates its own dedicated connection, which leads to hundreds of open connections over time.

It resolves the problem by introducing a shared pgxpool implemented using a singleton pattern. All retrievers now reuse the same connection pool instead of creating new individual connections.

**How to test:**

1. Configure multiple Postgres retrievers. (100+)
2. Verify that only a few connections to the database are created. (the pool may create multiple)
3. Confirm that feature flag evaluations still work as expected.
4. Monitor Postgres to ensure connection counts stay stable.

**Breaking changes:**

None expected. Retrievers now use a pooled connection, but the change is internal and transparent to users.

## Closes issue(s)

Resolve #4392 

## Checklist
- [x] I have implemented reconnecting logic
- [x] I have tested this code


thomaspoignant:
@gogbog thanks for starting this, Gemini said exactly what I was about to say _(probably in a better way)_.
Your implementation is great but it will not work if you are using 2 DB (which can be the case for isolated environment for example).

I also was asking myself if we want to offer a way to opt-out from using the pool and use direct connection?

gogbog:
@thomaspoignant I totally missed that use-case. We can still make the solution support multiple pools depending on how many databases the user needs to connect to - for example:
```
var (
    pools    = map[string]*pgxpool.Pool{}
    refCount = map[string]int{}
    mu       sync.Mutex
)
```

For the opting-out from using the pool and using a direct connection instead, we can definitely add support for it, but what would be the benefits? From what I know, the pool - even if you only ever use a single connection - is not a bad choice because it still manages reconnections, backoff, lifecycle, and error recovery for free, whereas a direct connection gives you none of that.

thomaspoignant:
Your proposal with a `Map` is great, this is what I had in mind too.

For the optout, forget it, you are right it will always be better to have a pool.

PR #4565, opened 2025-12-30
fix(relayproxy): fix flaky config loader callback test

thomaspoignant:
## Description

This fixes a flaky test in `TestConfigLoaderAddConfigChangeCallback` where multiple `fsnotify` events could trigger multiple callbacks for a single file change.

### Problem
When a file is modified, `fsnotify` can emit multiple events in rapid succession (e.g., `WRITE`, `CHMOD`, etc.). The previous implementation would process each event separately, causing the callback to be invoked multiple times for a single logical file change. This resulted in flaky tests that sometimes received more callback invocations than expected.

### Solution
The fix uses a channel-based event processing approach:
- File change events are sent to a buffered channel with size 1 (non-blocking send)
- A dedicated goroutine processes events from the channel
- Multiple rapid events are naturally coalesced since the buffered channel can only hold one pending event

This approach:
1. Decouples the `fsnotify` callback from the heavy config reloading work
2. Ensures rapid successive events result in only one config reload
3. Provides clean shutdown via a stop channel with `sync.Once` protection

### Testing
- Ran the test 10+ times locally with no failures
- The fix addresses the root cause rather than relying on timing-based workarounds

## Closes issue(s)
N/A - This is a test reliability fix

## Checklist
- [x] I have tested this code

> Documentation update not needed - this is an internal implementation change with no user-facing impact.

### DIFF
--- a/cmd/relayproxy/config/config_loader.go
+++ b/cmd/relayproxy/config/config_loader.go
@@ -101,6 +101,9 @@ func (c *ConfigLoader) startWatchChanges() {
 	}
 
 	// Start the event processor goroutine
+	// This approach is used to avoid blocking the fsnotify goroutine
+	// The event processor goroutine is responsible for processing the events and reloading the configuration
+	// Check https://github.com/knadh/koanf/issues/12#issuecomment-2637665148 to understand why this is necessary
 	go c.processConfigChangeEvents()
 
 	// The file watcher callback just sends to the channel and returns immediately


thomaspoignant:
### Additional Context

This fix is based on the recommendation from the koanf maintainer in this discussion:
https://github.com/knadh/koanf/issues/12#issuecomment-2637665148

The key insight is that if the callback is doing heavy work (like reloading configuration), the simplest solution is to "unburden the callback" by pushing events to a channel and returning immediately. A separate goroutine then processes the events, which naturally coalesces rapid successive events.

PR #4624, opened 2026-01-13
docs: Update Ruby SDK documentation for GoFeatureFlagProvider

olvlvl:
The example is invalid. It uses "import" instead of Ruby's "require".

## Description

The example doesn't compile.


thomaspoignant:
@olvlvl thanks for your PR, I will merge this PR.

PR #4713, opened 2026-01-28
fix: error when JsonLogic panics

### LINKED ISSUE #4711: (bug) Internal server error when JsonLogic package pancis
### Observed behaviour

When using JsonLogic based targeting rules an internal server error is occurs if the underlying package returns an error. For example take flag with the following JsonLogic as targeting rule query:
```json
{
          "==": [
            "end",
            {
              "substr": [
                {
                  "var": "my_short_var"
                },
                -3
              ]
            }
          ]
        }
```
When evaluating the flag and not providing the variable `my_short_var` at all in the context or providing a variable that is only `2` characters (shorter than the -3 substring), the JsonLogic package panics because the array index is out of bounds. This line specifically in the underlying package panics: https://github.com/GeorgeD19/json-logic-go/blob/master/jsonlogic.go#L310
This results in an internal server error returned:
```
$ curl -X POST localhost:1031/ofrep/v1/evaluate/flags/my_flag --data '{"context": {"my_short_var": "ab"}}' --header 'Content-Type: application/json' 
{"message":"Internal Server Error"}
```

### Expected Behavior

As far as I know generally JsonLogic shouldn't return an any errors, but will just consider the expression falsey. This is for example how the javascript implementation handles it. So either the underlying package is patched to make sure it does not panic (I'm not sure if there are any other cases in which the package panics) or the panic is caught within go feature flag and when evaluating the flag the targeting rule is simply considered `false`.

### Steps to reproduce

See above

doriandekoning:
## Description
JsonLogic panics, panic not caught resulting in an internalservererror returned. Made the fix a bit more general, it will also catch any panics from `nikunjy` if it should ever throw some. 

## Closes issue(s)

Resolve #4711 

## Checklist
- [ ] I have tested this code
](https://github.com/thomaspoignant/go-feature-flag/issues/4711)

### DIFF
--- a/modules/core/flag/rule.go
+++ b/modules/core/flag/rule.go
@@ -281,6 +281,10 @@ func (r *Rule) getPercentageBuckets() map[string]percentageBucket {
 // MergeRules is merging 2 rules.
 // It is used when we have to update a rule in a scheduled rollout.
 func (r *Rule) MergeRules(updatedRule Rule) {
+	if updatedRule.Disable != nil {
+		r.Disable = updatedRule.Disable
+	}
+
 	if updatedRule.Query != nil {
 		r.Query = updatedRule.Query
 	}
--- a/modules/core/flag/rule_test.go
+++ b/modules/core/flag/rule_test.go
@@ -648,6 +648,37 @@ func TestRule_MergeRules(t *testing.T) {
 			},
 		},
 		{
+			name: "merge rule with disable set to true",
+			originalRule: flag.Rule{
+				Name:            testconvert.String("rule1"),
+				VariationResult: testconvert.String("variation_A"),
+			},
+			updatedRule: flag.Rule{
+				Disable: testconvert.Bool(true),
+			},
+			want: flag.Rule{
+				Name:            testconvert.String("rule1"),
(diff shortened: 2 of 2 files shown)


modules/core/flag/rule.go
```diff
@@ -102,11 +102,23 @@ func (r *Rule) Evaluate(key string, ctx ffcontext.Context, flagName string, isDe
...
 	mapCtx := utils.ContextToMap(ctx)
+
+	// Catch any panics that may occur during evaluation
+	defer func() {
+		if r := recover(); r != nil {
+			slog.ErrorContext(context.Background(), "panic recovered while evaluating query",
+				slog.Any("panic", r),
+				slog.String("query", query),
+			)
+			success = false
```
thomaspoignant, line 118: do we have to rename the result?
Can't we just return false here?

doriandekoning: Yep, because it is a deferred function we cannot simply return the value because it will get discarded as far as I know.

thomaspoignant: Ok no worries, I was just checking.
I will merge this PR.

PR #4777, opened 2026-02-06
test: Add semver targeting validation tests

grambbledook:
## Description

**What was the problem?**
We needed to make sure semver comparison in targeting rules works correctly and catch regressions early.

**How is it resolved?**
Added a couple of test coverage for semver targeting rules, including:
- Integration tests in provider_test.go for semver targeting rule `version ge 10.0.0-10`
  - Validates `10.0.0-11` >= `10.0.0-10` (true)
  - Validates `10.0.0-2` >= `10.0.0-10` (false)
- Minor fixes in unit switching string comparison to semver comparrison.

**How can we test the change?**
Run the test suite:
```bash
make test

make provider-test
```

### DIFF
--- a/openfeature/provider_tests/integration_tests.sh
+++ b/openfeature/provider_tests/integration_tests.sh
@@ -27,7 +27,6 @@ RELAY_PROXY_PID_AUTHENTICATED=$!
 wait_relay_proxy 1031
 wait_relay_proxy 1032
 
-
 # Launch java integration tests
 echo "------------------------------------------------------------------------------------------------"
 echo "----------- JAVA PROVIDER TESTS ----------------------------------------------------------------"


grambbledook:
Lint check fails, but I see it fails in main since dependabot's update to go-aws-sdk.

openfeature/provider_tests/integration_tests.sh
```diff
@@ -27,6 +27,7 @@ RELAY_PROXY_PID_AUTHENTICATED=$!
 wait_relay_proxy 1031
 wait_relay_proxy 1032
 
+
```
thomaspoignant: **🤓 nitpick:** ‏we don't think we need this extra line.

PR #4796, opened 2026-02-11
chore: use json-logic-go package with fix for in array partial match bug

doriandekoning:
## Description
As discussed in slack:
We're encountering quite a critical bug, the in operator for JsonLogic rules also does partial matching, so this rules like this return true for example: {"in": ["a", ["aa"]}  which is confirmed by looking at [the code of the rules package](https://github.com/GeorgeD19/json-logic-go/blob/master/jsonlogic.go#L319C1-L320C1), but doesn't comply with the jsonlogic spec. I could open a PR to this package as the fix seems quite simple but it seems unmaintained  so it probably won't be merged. 

The test I added fails with non fixed version of the package but succeeds with the replace directive.

## Checklist
- [x] I have tested this code

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -123,6 +123,8 @@ linters:
         - github.com/nikunjy/rules
         # fix for Athena v3 parquet query issue - https://github.com/xitongsys/parquet-go/issues/547
         - github.com/xitongsys/parquet-go
+        # fix for in array partial match bug - https://github.com/GeorgeD19/json-logic-go/pull/65
+        - github.com/GeorgeD19/json-logic-go
   exclusions:
     paths:
       - (^|/)bin($|/)


modules/core/go.mod
```diff
@@ -12,6 +12,10 @@ require (
 // TODO: remove this once https://github.com/nikunjy/rules/pull/43 merges and a new version is available
 replace github.com/nikunjy/rules => github.com/hairyhenderson/rules v0.0.0-20250704181428-58ee76134adc
 
+// This change is needed to fix a bug with the jsonlogic IN operator also matching partial matches
+// Fixes this https://github.com/GeorgeD19/json-logic-go/pull/2 but not yet merged
+replace github.com/GeorgeD19/json-logic-go => github.com/doriandekoning/json-logic-go v0.0.0-20260211095634-9d88fadd8311
+
```
thomaspoignant, line 18: ⚠️ warning: We encounter a bug with the other replace and we should add a similar replace in the top level `go.mod`.

doriandekoning: Fixed!

PR #4804, opened 2026-02-11
fix: replace directive for json-logic-go package

doriandekoning:
## Closes issue(s)

Instead of using the replace which every user of the package will have to add too, this just uses a tagged version of the fork of json-logic-go.

When this is released we should bump the version in the root go.mod. 

## Checklist
- [x] I have tested this code

### DIFF
--- a/modules/core/flag/rule.go
+++ b/modules/core/flag/rule.go
@@ -8,7 +8,7 @@ import (
 	"sort"
 	"time"
 
-	jsonlogic "github.com/GeorgeD19/json-logic-go"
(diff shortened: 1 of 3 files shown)


thomaspoignant:
@doriandekoning I am looking to replace the jsonlogic lib by `github.com/diegoholiveira/jsonlogic/v3` and I have open a pull request for this #4805.

This will most likely help to fix the issue.
I have to test the `wasm` generation in all supported languages before releasing it.

I would appreciate if you can review the PR.

PR #4941, opened 2026-03-07
fix(python-provider): use X-API-Key header for relay API keys

### LINKED ISSUE #4319: (bug) API Key Implementation Not According to Spec
### Observed behaviour

The current implementation uses `Authorization: Bearer <api-key>` header for authorization. However, [the OFREP spec specifies 2 ways of authorization](https://openfeature.dev/docs/reference/other-technologies/ofrep/openapi#tag/OFREP-Core/operation/evaluateFlag):
1. API key auth with header `X-API-Key: <api-key>`
2. OAuth 2.0 bearer token auth with header `Authorization: Bearer <jwt-token>`

### Expected Behavior

We should parse `X-API-Key` header rather than `Authorization` header for api key. 

I believe the best way forward is:
1. Mark the use of `Authorization` header as deprecated, and update the openfeature providers to use the `X-API-Key` header instead of the `Authorization` header (a breaking change for the openfeature providers). 
2. Once we decide to start supporting OAuth 2.0 authorization, we should then make a breaking change to the proxy so that the `Authorization` header is treated as an OAuth JWT token. 

### Steps to reproduce

_No response_

SergioChan:
## Description
This updates the Python OpenFeature provider to send configured relay API keys with the OFREP-compliant `X-API-Key` header instead of `Authorization: Bearer ...`.

### What was the problem?
Issue #4319 tracks the migration from legacy `Authorization` API-key usage to `X-API-Key` so `Authorization` can be reserved for future OAuth tokens.

### How is it resolved?
- Switched evaluation requests to set `X-API-Key` when `api_key` is configured.
- Switched data-collector requests to set `X-API-Key` when `api_key` is configured.
- Added unit tests asserting both paths send `X-API-Key` and do not send `Authorization`.

### How can we test the change?
- `python3 -m py_compile openfeature/providers/python-provider/gofeatureflag_python_provider/provider.py openfeature/providers/python-provider/gofeatureflag_python_provider/data_collector_hook.py openfeature/providers/python-provider/tests/test_gofeatureflag_python_provider.py`
- Added tests:
  - `test_should_send_api_key_in_x_api_key_header_for_evaluation`
  - `test_should_send_api_key_in_x_api_key_header_for_data_collector`

> Note: full pytest execution was not runnable in this environment because `pytest`/`poetry` are not installed.

## Closes issue(s)
Resolve #4319

## Checklist
- [x] I have tested this code

### DIFF
--- a/openfeature/providers/python-provider/gofeatureflag_python_provider/data_collector_hook.py
+++ b/openfeature/providers/python-provider/gofeatureflag_python_provider/data_collector_hook.py
@@ -112,7 +112,7 @@ class DataCollectorHook(Hook):
                 )
                 headers = {"Content-Type": "application/json"}
                 if self._options.api_key:
-                    headers["Authorization"] = "Bearer {}".format(self._options.api_key)
+                    headers["X-API-Key"] = self._options.api_key
 
                 response = self._http_client.request(
                     method="POST",
--- a/openfeature/providers/python-provider/gofeatureflag_python_provider/provider.py
+++ b/openfeature/providers/python-provider/gofeatureflag_python_provider/provider.py
@@ -195,7 +195,7 @@ class GoFeatureFlagProvider(BaseModel, AbstractProvider, metaclass=CombinedMetac
             else:
                 headers = {"Content-Type": "application/json"}
                 if self.options.api_key is not None:
-                    headers["Authorization"] = "Bearer {}".format(self.options.api_key)
+                    headers["X-API-Key"] = self.options.api_key
                 url = "{}{}".format(
                     str(self.options.endpoint).rstrip("/"),
                     "/v1/feature/{}/eval".format(flag_key),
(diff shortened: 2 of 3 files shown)


thomaspoignant:
@SergioChan thanks for your pull request.
A new python provider is almost ready and is using the `X-API-Key` header.

You can check the PR #4861.

SergioChan:
Thanks for the context — great to hear the new provider is aligned on `X-API-Key`. This PR keeps the Python provider in step with that direction (and issue #4319). If you want any additional compatibility coverage before merge, I can add it.

SergioChan:
Thanks for the heads-up — that makes sense.

Given the incoming Python provider changes, I’m closing this PR to avoid churn/redundancy. If there’s still a gap after that lands, I can reopen with any needed follow-up adjustments.

SergioChan:
Closing as discussed to avoid overlap with the upcoming provider update.

PR #5026, opened 2026-03-25
feat: Ignore top-level keys started with `x-`

ilia-rassadin-rn:
Currently, go-feature-flag doesn't support reusable targeting conditions out of the box. So we cannot define set of rules

```yaml
my-custom-group:
  - query: role eq "admin"
  - query: group eq "QA"
  - query: email eq "custom@example.com"

...
my-flag:
  targeting:
    - query: my-custom-group eq true
      variation: enabled
```

But something similar can be achieved using YAML anchors. The issue with this approach is go-feature-flag treats every top-level key as a flag. So this is invalid configuration because my-custom-group is invalid flag:

```yaml
x-custom-group: &x-custom-group
  - query: role eq "admin"
    variation: enabled
  - query: group eq "QA"
    variation: enabled
  - query: email eq "custom@example.com"
    variation: enabled

...
my-flag:
  variations:
    enabled: true
    disabled: false
  targeting:
    - *x-custom-group
  defaultRule:
    variation: disabled
```

It is possible to define anchors as part of the flag and then reuse it, but it lowers readability of the config file.

Usually, YAML consumers ignores keys started with `x-`. E.g. docker compose does this since 2017, openapi allows `x-` keys for extending schema since it was swagger, etc.

So this commit brings this approach into `go-feature-flag`. Now we can define top-level anchors started with `x-` and reuse these snippets to define multiple flags.

I've decided to ignore `x-` keys for JSON and TOML too. E.g. for JSON it might be used as a comment, so there is still value provided. And it is nice to be able to transparently convert flags definition between formats so they remain valid.

## Closes issue(s)

Resolve #

## Checklist
- [ ] I have tested this code

### DIFF
--- a/cmd/cli/helper/read_config_file.go
+++ b/cmd/cli/helper/read_config_file.go
@@ -66,19 +66,22 @@ func readConfigFile(configFile, configFormat string) (map[string]dto.DTO, error)
 		if err != nil {
 			return nil, fmt.Errorf("%s: could not parse file (toml): %w", configFile, err)
 		}
-		return flags, nil
 	case "json":
 		err := json.Unmarshal(dat, &flags)
 		if err != nil {
 			return nil, fmt.Errorf("%s: could not parse file (json): %w", configFile, err)
 		}
-		return flags, nil
 	default:
 		// default is YAML
 		err := yaml.Unmarshal(dat, &flags)
 		if err != nil {
 			return nil, fmt.Errorf("%s: could not parse file (yaml): %w", configFile, err)
 		}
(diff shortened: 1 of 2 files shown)


thomaspoignant:
Sorry for the delay answering this.
I understand why you need this but I am doubting if `x-` is the right prefix to use, because some companies may already have flags starting with this pattern and it will become a breaking change.

I am wondering if the prefix should not be configurable instead of being default to `x-`.

What do you think @ilia-rassadin-rn ?

PR #5174, opened 2026-04-29
perf(rule): cache parsed nikunjy targeting query evaluators

Sha-x2-nk:
### What was the problem?
For nikunjy-format targeting queries (the default), `evaluateRule` in `modules/core/flag/rule.go` calls `parser.Evaluate(query, mapCtx)`. Under the hood that constructs a fresh `parser.Evaluator` on every invocation, which runs the full ANTLR lex + parse over the query string. For a flag with N targeting rules, that's N ANTLR parses per evaluation, dominating both CPU and allocations.

CPU/alloc profiling on a flag with ~13 targeting rules showed roughly **2,000+ allocations per evaluation**, with the alloc graph dominated by `antlr4-go/antlr/v4.NewCommonToken`, `antlr4-go/antlr/v4.(*BaseLexer).Emit`, `nikunjy/rules/parser.NewQueryContext`, and similar parser/lexer constructor calls.

### How is it resolved?
Cache one `*parser.Evaluator` per distinct query string in a package-level `sync.Map`. Because `parser.Evaluator.Process` writes to an internal `lastDebugErr` field, sharing a single instance across goroutines would trigger Go's race detector. Each cache entry instead holds a small `sync.Pool` of `*parser.Evaluator` for the same query, so:

- Parsing is amortized — a query is parsed once (or a small constant number of times under contention) regardless of evaluation count.
- Concurrent callers don't serialize on a mutex.
- No behavioral change to `Rule.Evaluate` or any public API.

The cache is keyed only by the query string. `parser.NewEvaluator` is deterministic in its input, so this is safe. Cache entries are not evicted; in practice the set of distinct query strings is bounded by the flag config and is much smaller than the request volume.

### How can we test the change?
Existing flag-package tests cover correctness (`go test -race ./modules/core/flag/...` passes locally).

A new file `modules/core/flag/rule_bench_test.go` adds two benchmarks. Measured on darwin/arm64 (Apple M2 Pro), Go 1.25:

**6-predicate query** (`language eq "ar" and isNewUser eq true and clubsTimeSpent gt 600 and clubsTimeSpent le 3600 and concurrencyLocked eq true and segment eq 0`):

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| main | 19,608 | 21,344 | 301 |
| this PR | 2,105 | 1,198 | 36 |
| speedup | **9.3×** | 18× | 8× |

**Single-predicate query** (`language eq "ar"`):

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| main | 3,628 | 5,192 | 85 |
| this PR | 787 | 865 | 19 |
| speedup | **4.6×** | 6× | 4.5× |

The improvement scales with the size of the query and the number of rules per flag.

### Breaking changes
None. Public API unchanged; behavior identical.

## Closes issue(s)
N/A (filing this directly as a perf improvement; happy to file a tracking issue if preferred).

## Checklist
- [x] I have tested this code (`go test -race ./modules/core/flag/...`, `make lint`)
- [x] I have added unit test to cover this code (benchmark file; existing rule tests cover correctness)
- [ ] I have updated the documentation (`README.md` and `/website/docs`) — not applicable, internal-only change


modules/core/flag/rule.go
```diff
@@ -17,6 +18,60 @@ import (
...
 
+// nikunjyEvaluatorCache memoizes a sync.Pool of *parser.Evaluator per query string.
+// Parsing the query via ANTLR is the dominant cost of a nikunjy rule evaluation
+// (see https://github.com/nikunjy/rules/blob/master/parser/evaluate.go); the parsed
+// Evaluator is reusable, but parser.Evaluator.Process writes to an internal field, so
+// concurrent calls require either serialization or a pool of evaluators per query.
+// We use sync.Pool because evaluations are typically high-throughput and short-lived.
+//
+// Memory note: cache entries are not evicted, since a feature flag's query strings
+// are part of static config and the set is bounded in practice.
```
thomaspoignant, line 29: **❔ question:** ‏I am worried about this, since we are able to change to configuration on run time, we will have flag config edits add new query strings.

It means that this is dynamic and it can grow at every configuration changes.
I am worried that long-lived process with many config reloads leaks.

I am curious to have your point of view on this?

thomaspoignant (commented):
@Sha-x2-nk thanks a ton for this pull request, the performance gain is huge.
I am just worried about not having any limits in the cache, because every config change will add on top and it can leak.

What about using a LRU cache instead here?

thomaspoignant:
I will merge this one and we will implement the LRU Cache if needed in future version.

thomaspoignant (approved):

PR #5322, opened 2026-05-23
feat: fix json retriever format hint

### LINKED ISSUE #5321: (change) Redis retriever: default FileFormat causes unnecessary YAML parsing of JSON data
### Motivation

The Redis, MongoDB, and PostgreSQL retrievers internally serialize flag data using `json.Marshal`, but the library defaults `FileFormat` to YAML. Since JSON is valid YAML, parsing succeeds — but YAML parsing is 2–3x slower than JSON parsing, causing unnecessary CPU overhead.

The `ConvertToFlagStruct` function selects the parser based on a single global `FileFormat` config:

```go
// internal/cache/cache_manager.go:54-70
func ConvertToFlagStruct(loadedFlags []byte, fileFormat string) (map[string]dto.DTO, error) {
    switch strings.ToLower(fileFormat) {
    case "toml":
        err = toml.Unmarshal(loadedFlags, &newFlags)
    case "json":
        err = json.Unmarshal(loadedFlags, &newFlags)
    default:
        // default unmarshaller is YAML
        err = yaml.Unmarshal(loadedFlags, &newFlags)
    }
}
```

When `FileFormat` is empty (the default), all retriever output is parsed as YAML — even when the retriever always produces JSON.
In my case, using redis, all flags are parsed as YAML. 

The same `FileFormat` is applied uniformly to all retrievers in `retriever/manager.go:313`:
```go
convertedFlag, err := cache.ConvertToFlagStruct(rawValue, format)
```

These retrievers hardcode `json.Marshal` internally and always return JSON bytes:

| Retriever | File | Line |
|-----------|------|------|
| Redis | `retriever/redisretriever/retriever.go` | 97 |
| MongoDB | `retriever/mongodbretriever/retriever.go` | 94 |
| PostgreSQL | `retriever/postgresqlretriever/retriever.go` | 133 |

File-based retrievers (File, HTTP, S3, GCS, Azure Blob, GitHub, GitLab, Bitbucket, K8s) are unaffected since they return raw bytes in whatever format the user actually stores.

### Requirements

I would like to update retrievers to declare their output format via an optional interface:

```go
// retriever/retriever.go
type FormatHinter interface {
    OutputFormat() string // "json", "yaml", "toml"
}
```

Then Redis, MongoDB, and PostgreSQL retrievers would implement `OutputFormat() string { return "json" }`.

hugehoo:
## Description
This PR lets `retrievers` that always emit JSON (Redis, MongoDB, PostgreSQL) opt out of the default YAML parsing path, removing unnecessary CPU overhead from feature flag evaluation.

### What was the problem?

- The Redis, MongoDB, and PostgreSQL retrievers serialize their output with `json.Marshal` before returning it from `Retrieve()`.
- However, the library defaults `FileFormat` to YAML. With no explicit `FileFormat` set, `cache.ConvertToFlagStruct` falls into the `default` branch of its switch and parses every retriever's bytes with `yaml.Unmarshal`.
- JSON is a valid subset of YAML, so parsing succeeds — but **YAML parsing is 2–3x slower than JSON parsing**, which translates to roughly ~4% extra CPU in production for users on these retrievers.
- The current design applies the same global `FileFormat` to every retriever; there is no way for a retriever that inherently produces JSON to signal that fact to the parser.

### How is it resolved?

Introduce an **optional `FormatHinter` interface** that a retriever can implement to declare the format of the bytes it returns:

```go
type FormatHinter interface {
    OutputFormat() string // "json", "yaml", or "toml"
}
```

The retrieve manager checks for this interface per-retriever and uses the declared format when present, falling back to the global `FileFormat` otherwise. This keeps the change fully backwards-compatible — retrievers that do not implement `FormatHinter` behave exactly as before.

## Closes issue(s)

Resolve #5321

## Checklist
- [x] I have tested this code


thomaspoignant (approved):
Thanks a lot @hugehoo this is really helpful and it will help performance wise.

PR #5356, opened 2026-06-03
feat: add Google BigQuery exporter

mvanhorn:
## Summary

Adds a first-class Google BigQuery exporter that streams flag-evaluation and tracking events directly into BigQuery using the official `cloud.google.com/go/bigquery` streaming-insert API (`Inserter.Put()`), with no GCS intermediate stage.

- New `exporter/bigqueryexporter` package implementing the bulk `Exporter` interface (`Export` + `IsBulk()` returning `true`), modeled on `pubsubexporterv2`.
- Routes events by concrete type: `FeatureEvent` -> `feature_flag_evaluations`, `TrackingEvent` -> `tracking_events`, using the column schemas from the issue. JSON columns (`value`, `metadata`, `tracking_details`, `evaluation_context`) are JSON-marshaled.
- Optional `AutoMigrate` creates the dataset and the tables being written if missing (existing ones treated as success).
- Supports Application Default Credentials, or an explicit `GoogleCredentials` JSON.
- Relay-proxy config: new `bigquery` kind in the `ExporterKind` enum + `IsValid()`, new `datasetID` / `tableName` / `trackingTableName` / `autoMigrate` fields (reusing the existing `projectID`), validation requiring `projectID` + `datasetID`, and a `case config.BigQueryExporter` branch in `createExporter`.
- Streaming inserts are chunked under BigQuery's per-request limits, and each row carries a deterministic `insertID` so retried requests are best-effort de-duplicated.
- New docs page under `website/docs/integrations/export-evaluation-data/`, plus the registry entry in `website/data/integrations.js`.

## Why this matters

Issue #5087 (owner-authored, fully specced) asks for a first-class BigQuery exporter so analytics/BI teams can join feature-flag data with application metrics already in BigQuery using standard SQL, without standing up and reprocessing a GCS export. The issue ships the exporter shape, both table schemas, the relay-proxy YAML, and acceptance criteria; this PR implements them directly.

## Testing

- `go build ./exporter/bigqueryexporter/... ./cmd/relayproxy/...` — clean.
- `go vet ./exporter/bigqueryexporter/...` — clean.
- `gofmt -l` on touched files — clean.
- `go test ./exporter/bigqueryexporter/... ./cmd/relayproxy/config/... ./cmd/relayproxy/service/...` — pass.
- New table-driven tests mock the BigQuery client/inserter seam (no network): both event types map to the documented columns and route to the correct table, mixed batches split per table in one `Put` each, `AutoMigrate` ensures only the tables being written, default table names apply, and `IsBulk()` returns `true`.
- `go mod tidy` run after adding the `cloud.google.com/go/bigquery` dependency.

Fixes #5087

AI was used for assistance.


exporter/bigqueryexporter/exporter.go
```diff
@@ -0,0 +1,381 @@
...
+	// DatasetID is the BigQuery dataset receiving the events.
+	DatasetID string
+
+	// TableName is the table receiving feature flag evaluation events.
+	// Default: feature_flag_evaluations
+	TableName string
+
+	// TrackingTableName is the table receiving tracking events.
+	// Default: tracking_events
+	TrackingTableName string
```
thomaspoignant: Why do we need both `TableName` and `TrackingTableName` here?
The way exporters works they are a different instance of the struct for tracking and evaluation.

This means that an exporter manage only 1 type at a time, so we don't a field for each table and `TableName` is enough here.

exporter/bigqueryexporter/exporter.go
```diff
@@ -0,0 +1,381 @@
...
+	featureRows, trackingRows, err := e.buildRows(events)
+	if err != nil {
+		return err
+	}
+	if len(featureRows) == 0 && len(trackingRows) == 0 {
+		return nil
+	}
+
+	if e.client == nil {
+		if err := e.initClient(ctx); err != nil {
```
thomaspoignant: We can have a data race and init the client multiple time with this.
Maybe a `sync.Once` will be better here to ensure that we don't have concurrency problems.

thomaspoignant (changes requested):
Thanks @mvanhorn for your contribution, I have made some feedbacks.
Let me know if you need any help.

mvanhorn:
Thanks for the review! Addressed in 5d6b101:

- You were right about the table fields - removed `TrackingTableName` entirely. Each exporter instance now writes to a single `TableName` and picks its row schema from the event type it receives, matching the one-instance-per-event-type model.
- Client init now uses `sync.Once` (with the init error stored), so concurrent Export calls can't double-init. Also dropped the `applyDefaults` mutation in favor of a read-only `tableName()` accessor so there's no shared-state write at all.
- While reworking the relay-proxy wiring I also added a `googleCredentials` passthrough to `ExporterConf` - without it the relay proxy could only use ADC even though the exporter supports explicit credentials JSON.
- Docs updated for the single-table model (two exporter blocks, one per eventType).

`go test -race` on the exporter package and the relay-proxy config/service suites pass locally.

thomaspoignant (approved):
@mvanhorn sorry for all the noise on automatic reviews on this PR, we are changing tool right now this is the reason why we had so many.

I've tested the PR and it works well, thanks for that.
I have just made a few changes on the docs (especially around the logo), I will now merge this PR.

PR #5478, opened 2026-06-21
feat: support PostgreSQL connection pool configuration

### LINKED ISSUE #4835: (change) PostgreSQL Connection Pool Configuration
### Motivation

Currently the PostgreSQL retriever supports basic connection settings such as the `uri`, `table`, and custom column names. However, it doesn’t provide user-configurable settings for connection pooling (max connections, idle connections, connection lifetimes, timeouts, etc.).

For many production workloads, especially under high concurrency or when using managed cloud Postgres instances, having control over these pool settings is essential for performance, stability, and resource efficiency.

### Requirements

Allow users to configure PostgreSQL connection pool parameters when using the postgresql retriever, instead of relying solely on driver defaults.

The configuration must:

1. Add an optional pool section under the PostgreSQL retriever configuration.
2. Map configuration values to the underlying Go database/sql pool settings - (`MaxOpenConns`, `MaxIdleConns`, `ConnMaxLifetime`, `ConnMaxIdleTime`)
3. If pool is not defined, current behavior and defaults must remain unchanged.
4. Validate configuration values

**Proposed solution:**

Introduce optional PostgreSQL retriever configuration properties for pool management. For example:

```yaml
retrievers:
  - kind: postgresql
    uri: "postgres://user:pass@host/db"
    table: "go_feature_flag"
    pool:
      max_open_conns: 50
      max_idle_conns: 10
      conn_max_lifetime: "30m"
      conn_max_idle_time: "10m"
```

mvanhorn:
## Description

The PostgreSQL retriever previously exposed only `uri`, `table`, and custom
column names, leaving the connection pool entirely at the driver defaults.
Production workloads on managed Postgres need control over pool sizing and
connection lifetimes for performance and resource safety.

This PR adds optional, per-retriever connection pool settings, following the
design the maintainer selected in #4835:

- `maxOpenConns` -> pgxpool `MaxConns`
- `maxIdleConns` -> pgxpool `MinConns` (the pool floor; pgxpool has no separate
  idle ceiling, idle connections are bounded by `MaxConns`)
- `connMaxLifetime` -> pgxpool `MaxConnLifetime`
- `connMaxIdleTime` -> pgxpool `MaxConnIdleTime`

How it works:

- A new `PoolConfig` is threaded through the `pgxpool` pool builder at the
  existing `GetPool` / `poolMap` seam in `postgres.go`. When no pool block is
  configured, the pool is built exactly as before (`pgxpool.New`), so existing
  configurations are byte-for-byte unaffected. When a pool block is present, the
  URI is parsed with `pgxpool.ParseConfig` and only the explicitly set fields are
  overridden.
- The cached `poolMap` is now keyed by URI **and** the pool config, so two
  retrievers that share a URI but request different pool settings no longer
  silently reuse the first pool created.
- The same fields are surfaced on `RetrieverConf` (relay proxy / config files)
  with validation in `validatePostgreSQLRetriever()` (negative values,
  `maxIdleConns` greater than `maxOpenConns`, and unparseable durations are
  rejected with a clear error) and wired through `createPostgreSQLRetriever`.
- Documentation for the new block is added to the PostgreSQL retriever docs.

How to test: see the new unit tests in
`retriever/postgresqlretriever/postgres_pool_test.go`,
`cmdhelpers/retrieverconf/retriever_conf_test.go`, and
`cmdhelpers/retrieverconf/init/retriever_init_test.go`. Building a pool does not
open a connection, so the pool-config tests run without a live database; the
existing testcontainers-backed pool test continues to cover end-to-end behavior.

No breaking change to configuration behavior: every new field is optional and
defaults preserve the current behavior.

## Closes issue(s)

Resolve #4835

## Checklist
- [x] I have tested this code

Fixes #4835

### DIFF
--- a/retriever/postgresqlretriever/postgres.go
+++ b/retriever/postgresqlretriever/postgres.go
@@ -14,30 +14,26 @@ import (
 // the corresponding pgxpool default (or the value already encoded in the
 // connection URI) is used, so existing configurations are unaffected.
 type PoolConfig struct {
-	// MaxOpenConns maps to pgxpool's MaxConns: the maximum number of
-	// connections in the pool.
-	MaxOpenConns int32
-	// MaxIdleConns maps to pgxpool's MinConns: the minimum number of
-	// connections the pool keeps open. pgxpool has no direct cap on idle
-	// connections (idle connections are bounded by MaxConns), so this controls
-	// the pool floor rather than an idle ceiling. It must not exceed
-	// MaxOpenConns.
-	MaxIdleConns int32
-	// ConnMaxLifetime maps to pgxpool's MaxConnLifetime: the duration since
-	// creation after which a connection is automatically closed.
-	ConnMaxLifetime time.Duration
-	// ConnMaxIdleTime maps to pgxpool's MaxConnIdleTime: the duration after
-	// which an idle connection is automatically closed by the health check.
-	ConnMaxIdleTime time.Duration
+	// MaxConns is the maximum number of connections in the pool.
+	MaxConns int32
+	// MinConns is the minimum number of connections the pool keeps open. It
+	// must not exceed MaxConns when both are configured.
+	MinConns int32
+	// MaxConnLifetime is the duration since creation after which a connection
+	// is automatically closed.
+	MaxConnLifetime time.Duration
+	// MaxConnIdleTime is the duration after which an idle connection is
+	// automatically closed by the health check.
+	MaxConnIdleTime time.Duration
 }
 
 // IsZero reports whether no pool setting has been configured. When true,
 // GetPool builds the pool exactly as it did before pool settings existed.
 func (p PoolConfig) IsZero() bool {
-	return p.MaxOpenConns == 0 &&
-		p.MaxIdleConns == 0 &&
-		p.ConnMaxLifetime == 0 &&
-		p.ConnMaxIdleTime == 0
+	return p.MaxConns == 0 &&
+		p.MinConns == 0 &&
+		p.MaxConnLifetime == 0 &&
+		p.MaxConnIdleTime == 0
 }
 
 // cacheKey returns a stable cache key for a URI + pool config combination.
@@ -47,8 +43,8 @@ func cacheKey(uri string, p PoolConfig) string {
 	if p.IsZero() {
 		return uri
 	}
-	return fmt.Sprintf("%s|%d|%d|%d|%d",
-		uri, p.MaxOpenConns, p.MaxIdleConns, p.ConnMaxLifetime, p.ConnMaxIdleTime)
+	return fmt.Sprintf("%s|maxConns=%d|minConns=%d|maxConnLifetime=%d|maxConnIdleTime=%d",
+		uri, p.MaxConns, p.MinConns, p.MaxConnLifetime, p.MaxConnIdleTime)
 }
 
 type poolEntry struct {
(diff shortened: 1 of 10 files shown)


mvanhorn:
Fixed both CI failures:

- **Lint (gosec G115):** the `int -> int32` conversions for `MaxOpenConns`/`MaxIdleConns` now have a bounds check that returns an error for out-of-range values before narrowing, so the overflow warning is resolved.
- **Test (check-mode goldens):** the four optional pool fields were serializing zero values into every retriever's config output, breaking the `check-*` golden comparisons. Added `json:",omitempty"` so they only appear when actually configured. All `TestCmdEvaluate` check-mode tests pass locally and gosec reports no G115 issues.

retriever/postgresqlretriever/postgres.go
```diff
@@ -38,27 +88,61 @@ func GetPool(ctx context.Context, uri string) (*pgxpool.Pool, error) {
...
+	config, err := pgxpool.ParseConfig(uri)
+	if err != nil {
+		return nil, err
+	}
+	if poolCfg.MaxOpenConns > 0 {
+		config.MaxConns = poolCfg.MaxOpenConns
+	}
+	if poolCfg.MaxIdleConns > 0 {
+		config.MinConns = poolCfg.MaxIdleConns
+	}
```
thomaspoignant: **🔨 issue:** ‏why do you put `MaxIdleConns` inside the `MinConns` pgxpool config here ?

Should it me 
```go
config.MaxIdleConns = poolCfg.MaxIdleConns
```

retriever/postgresqlretriever/postgres.go
```diff
@@ -38,27 +88,61 @@ func GetPool(ctx context.Context, uri string) (*pgxpool.Pool, error) {
...
+	if poolCfg.ConnMaxLifetime > 0 {
+		config.MaxConnLifetime = poolCfg.ConnMaxLifetime
+	}
+	if poolCfg.ConnMaxIdleTime > 0 {
+		config.MaxConnIdleTime = poolCfg.ConnMaxIdleTime
+	}
+
+	return pgxpool.NewWithConfig(ctx, config)
```
thomaspoignant, line 125: It would be great here to have a check that we don't have `MinConns` > `MaxConns`

Something like this would work:


```suggestion
  if config.MinConns > config.MaxConns {
    return nil, fmt.Errorf(
        "invalid pool configuration: maxIdleConns (%d) must not exceed maxOp
enConns (%d)",
        config.MinConns, config.MaxConns)
  }

  return pgxpool.NewWithConfig(ctx, config)
```

retriever/postgresqlretriever/postgres.go
```diff
@@ -38,27 +88,61 @@ func GetPool(ctx context.Context, uri string) (*pgxpool.Pool, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	if poolCfg.MaxOpenConns > 0 {
+		config.MaxConns = poolCfg.MaxOpenConns
+	}
+	if poolCfg.MaxIdleConns > 0 {
+		config.MinConns = poolCfg.MaxIdleConns
+	}
+	if poolCfg.ConnMaxLifetime > 0 {
```
thomaspoignant: **🤓 nitpick:** ‏If we want to test fully.
```suggestion
	if c.MaxOpenConns < 0 || c.MaxOpenConns > math.MaxInt32 {
```

retriever/postgresqlretriever/postgres.go
```diff
@@ -38,27 +88,61 @@ func GetPool(ctx context.Context, uri string) (*pgxpool.Pool, error) {
...
+		config.MaxConnLifetime = poolCfg.ConnMaxLifetime
+	}
+	if poolCfg.ConnMaxIdleTime > 0 {
```
thomaspoignant: **🤓 nitpick:** ‏very very nit pick
```suggestion
	if poolCfg.ConnMaxIdleTime > 0  || c.ConnMaxIdleTime > math.MaxInt32 {
```

retriever/postgresqlretriever/postgres.go
```diff
@@ -38,27 +88,61 @@ func GetPool(ctx context.Context, uri string) (*pgxpool.Pool, error) {
...
+	if poolCfg.IsZero() {
+		return pgxpool.New(ctx, uri)
+	}
+
+	config, err := pgxpool.ParseConfig(uri)
+	if err != nil {
+		return nil, err
+	}
+	if poolCfg.MaxOpenConns > 0 {
+		config.MaxConns = poolCfg.MaxOpenConns
```
thomaspoignant: 🔨 issue: ‏why do you put MaxOpenConns inside the MaxConns pgxpool config here ?

Should it be?
```go
config.MaxOpenConns = poolCfg.MaxOpenConns
```

thomaspoignant (commented):
Hey @mvanhorn I've done a first path of reviews (not complete yet), but I have addressed some small issues here and there.

I will perform a more in depth review as soon as possible.

mvanhorn:
Good catch on the mapping - you're right that borrowing database/sql's MaxOpenConns/MaxIdleConns names was misleading, since MaxIdleConns (a max) was landing on pgxpool's MinConns (a minimum floor). Renamed the pool config to pgxpool's own model: maxConns/minConns, mapped 1:1 (config.MaxConns = poolCfg.MaxConns, config.MinConns = poolCfg.MinConns). Updated the yaml/json keys, docs, and tests to match.

Also applied your suggestions: a MinConns > MaxConns validation (both at the retriever config layer and in newPool), and fixed the int32 bounds checks to < 0 || > math.MaxInt32 for both fields.

Verified: the pool config tests pass locally (IsZero, AppliesConfig, the MinConns>MaxConns guard, partial config, invalid URI) and go vet is clean. The one failing test, TestGetPool_MultipleURIsAndReuse, spins up a real Postgres via testcontainers and needs Docker running, which I don't have here - it's unaffected by the rename.

mvanhorn:
The Govulncheck failure is GO-2026-5856, a Go stdlib (crypto/tls) vulnerability fixed in go1.26.5 - it's unrelated to this PR's diff and will affect any branch on go1.26.4. Bumped the go directive to 1.26.5 and re-ran make's vuln-check target locally: 0 affecting vulnerabilities now. If you'd rather take the bump on main separately, happy to drop the commit from this PR.

PR #5676, opened 2026-07-23
fix(data-exporter): avoid copying EventStore with embedded sync.RWMutex

AruneshDwivedi:
The Flush method in data_exporter.go was doing store := *d.eventStore which copies
the eventStoreImpl struct, including its embedded sync.RWMutex.

Copying a struct with an embedded mutex is undefined behavior in Go - the copy gets
a zero-value mutex that has never been initialized, so all lock/unlock operations
on it do nothing and don't actually protect anything.

Changed to use pointer directly (store := d.eventStore) so ProcessPendingEvents
operates on the original mutex-protected store.


thomaspoignant:
@AruneshDwivedi thanks for your PR, but it is not compiling anymore (check the CI output).
Are you sure about your fix?

thomaspoignant:
Thanks for the PR @AruneshDwivedi, and for digging into the exporter internals — removing the pointer-to-interface here is a real improvement and I'd like to merge it. But I need to correct the diagnosis before it goes in, because the commit message and PR description describe a bug that isn't there.

**`EventStore[T]` is an interface, not a struct.**

```go
// exporter/event_store.go:49
type EventStore[T ExportableEvent] interface { ... }
```

The `sync.RWMutex` lives in the unexported `eventStoreImpl[T]` struct (`event_store.go:13`), and `NewEventStore` returns `&eventStoreImpl[T]{...}` — a pointer. So the old code:

```go
store := *d.eventStore   // d.eventStore was *EventStore[T]
```

was dereferencing a *pointer to an interface*. That copies the two-word interface header (type pointer + data pointer), not the struct behind it. The `eventStoreImpl` — and its mutex — was never copied. Every consumer was already sharing the same store and the same lock.

Two checks confirm it on the base commit, before your change:

- `go vet ./exporter/` is clean. The `copylocks` analyzer exists precisely to catch copying a value containing a `sync.Locker`, and it fires on nothing here.
- `go test -race -count=3 -run TestDataExporterFlush ./exporter/` passes.

So there was no undefined behavior and no unprotected critical section to fix.

**That said, the change itself is right.** `*EventStore[T]` is a pointer-to-interface, which is nearly always a Go code smell: it adds an indirection that does nothing, breaks method-set resolution, and invites exactly this kind of misreading. Dropping it to plain `EventStore[T]` is the correct shape. I'd just like it landed as what it is.

Before merging:

1. **Retitle and rewrite the description.** `refactor(exporter): drop pointer-to-interface for EventStore`, with a body saying it removes a redundant indirection and has no behavior change. As written, `fix(data-exporter):` will cut a patch release whose changelog advertises a concurrency fix to users who never had one — that's the part I most want to avoid.

2. **Finish the same cleanup in `manager.go`,** which the diff already touches. `managerImpl` still carries the identical pattern:

   ```go
   eventStore *EventStore[T]   // manager.go:21
   eventStore: &evStore,       // manager.go:42
   store := *m.eventStore      // manager.go:47, in AddEvent
   ```

Both are unexported, so it's zero-risk, and leaving it means the codebase ends up with two conventions for the same field.

No new test needed — `TestDataExporterFlush_ConcurrentFlushNoDuplicate` passes identically before and after, since it exercises the per-consumer `mutex` in `ProcessPendingEvents` (`event_store.go:106`), which this doesn't touch. There's no regression to guard against.

One unrelated thing I noticed while reading: `managerImpl.Stop()` (`manager.go:69`) stops each consumer but never calls `m.eventStore.Stop()`, so the `periodicCleanQueue` goroutine from `event_store.go:37` leaks for the process lifetime. Not yours — I'll open a separate issue.

thomaspoignant (approved):

PR #5903, opened 2026-08-25
feat(relayproxy): advertise the SSE endpoint in the OFREP bulk evaluation response

### LINKED ISSUE #5675: (feature) relayproxy -> implement OpenFeature ADR0008 "Server-Sent Events (SSE) for bulk evaluation changes"
### Requirements

**Current behaviour**
The relayproxy seems to not include the `eventStreams` field into the [OFREPBulkEvaluationSuccess](https://github.com/thomaspoignant/go-feature-flag/blob/main/cmd/relayproxy/model/ofrep_bulk_success_response.go) response model when using OFREP endpoint `ofrep/v1/evaluate/flags`.

**Expected behaviour**
The relayproxy should implement [OpenFeature ADR0008](https://github.com/open-feature/protocol/blob/main/service/adrs/0008-sse-for-bulk-evaluation-changes.md) to let OFREP providers use the `eventStreams` field and eventually read for flag changes, without the needs of polling.

**Proposals and Considerations**
The relayproxy already expose am SSE endpoint `/stream/v1/sse/flag/change` that can be included in BulkEvaluationSuccess responses, for example:
`{ "flags": [...], "eventStreams": [{ "type": "sse", "url": "${RESOLVED_SSE_HOST}/stream/v1/sse/flag/change", "inactivityDelaySec": ${RESOLVED_INACTIVITY_DELAY} } ] }`

`RESOLVED_SSE_HOST` may be:
- resolved automatically from the incoming request, by using an internal context property in the payload like `gofeatureflag.sseEndpoint` or some request header like `x-gofeatureflag-sse-endpoint`
- resolved from a mapping configuration in the config file. For example a new `sse.endpoint` or `eventStream.endpoint` property.

The same may be applied to `RESOLVED_INACTIVITY_DELAY`.

NOTE: The feature may be initially considered opt-in through a flag in the config file like `sse.ofrepEnabled` or something else that fits better. This will let unchanged config files to make relayproxy to behave as-is.

**Eventual work to be done**
- Update the codebase of relayproxy
- Update documentation (configuration, usage, use cases, etc.)
  - Probably a dedicated section for [AWS Lambda streaming responses](https://docs.aws.amazon.com/lambda/latest/dg/configuration-response-streaming.html) + eventual use case when integrating with [AWS API Gateway streaming responses](https://docs.aws.amazon.com/apigateway/latest/developerguide/response-transfer-mode-lambda.html)

ChrisJr404:
## Description

The relay proxy already exposes an SSE endpoint for flag changes (`/stream/v1/sse/flag/change`), but the OFREP bulk evaluation response never told providers about it, so they had to keep polling.

This adds the `eventStreams` field to the OFREP bulk evaluation response (`/ofrep/v1/evaluate/flags`) as described in [OpenFeature ADR-0008](https://github.com/open-feature/protocol/blob/main/service/adrs/0008-sse-for-bulk-evaluation-changes.md). When enabled, the response points providers at the SSE endpoint so they can subscribe instead of polling.

As discussed in the issue, adding the field unconditionally would be a change for existing clients, so it is opt-in through a new `ofrepEventStream` config block and stays off by default. When disabled (or when no endpoint is set) the response is exactly the same as before.

```yaml
ofrepEventStream:
  enabled: true
  endpoint: "https://gofeatureflag.example.com"
  inactivityDelaySec: 60
```

The advertised URL is built from the configured endpoint, the client stays responsible for adding its own credentials when it connects.

### How to test

`go test ./cmd/relayproxy/handler/ofrep/...` covers the enabled case, the disabled default, and the enabled-without-endpoint case.

## Closes issue(s)

Resolve #5675

## Checklist
- [x] I have tested this code


cmd/relayproxy/config/config.go
```diff
@@ -20,6 +20,11 @@ type Config struct {
 	// Swagger is the swagger configuration
 	Swagger Swagger `mapstructure:"swagger" koanf:"swagger"`
 
+	// OfrepEventStream (optional) advertises the flag-change SSE endpoint in the OFREP bulk
+	// evaluation response so that providers can subscribe to changes instead of polling.
+	// It is disabled by default, so the bulk evaluation response is left unchanged.
+	OfrepEventStream OfrepEventStream `mapstructure:"ofrepEventStream" koanf:"ofrepeventstream"`
```
thomaspoignant, line 26: I wonder if we should not activate it by default because it becomes part of the protocol and we support it.
What do you think about this?

This may required a test of all the OFREP providers but I would prefer to do so.

cmd/relayproxy/handler/ofrep/evaluate.go
```diff
@@ -218,10 +226,28 @@ func (h *EvaluateCtrl) BulkEvaluate(c echo.Context) error {
...
+// ofrepSSEPath is the path of the relay proxy endpoint streaming flag changes over SSE.
+const ofrepSSEPath = "/stream/v1/sse/flag/change"
+
+// buildEventStreams returns the eventStreams advertised in the OFREP bulk evaluation
+// response. The endpoint comes from the configuration, the client is responsible for
+// adding its own credentials when connecting.
+func (h *EvaluateCtrl) buildEventStreams() []model.OFREPEventStream {
+	return []model.OFREPEventStream{{
+		Type:               "sse",
+		URL:                strings.TrimRight(h.eventStream.Endpoint, "/") + ofrepSSEPath,
```
thomaspoignant, line 246: In some cases there is a `apiKey` needed to call the SSE endpoint.
We need to be able to provide it when answering to the OFREP provider.

As said in the ADR:
> url (string, optional): The endpoint URL. This is the default representation and is opaque to the provider. It may include authentication tokens, channel identifiers, or other vendor-specific query parameters. Implementations must treat this URL as sensitive -- it may contain auth tokens or channel credentials -- and must not log or persist the full URL including query string.

So it is fine to return it directly.

In this case it will be set inside the query param `apiKey` as we check it in the [stream authorizer](https://github.com/thomaspoignant/go-feature-flag/blob/main/cmd/relayproxy/api/middleware/stream_authorizer.go).

cmd/relayproxy/config/config.go
thomaspoignant, line 26: There is no validation of this config inside [`config/config_validator.go`](https://github.com/thomaspoignant/go-feature-flag/blob/0a8179ecc4c6287595023735b0f45190cdfd5cb6/cmd/relayproxy/config/config_validator.go#L13), it may be great to fail explicitly if
- `enabled: true` with an empty endpoint silently no-ops.
- `endpoint` isn’t an absolute URL — a scheme-less value produces a relative URL that providers resolve against their own origin.
- etc ...

cmd/relayproxy/config/config_ofrep_event_stream.go
```diff
@@ -0,0 +1,17 @@
...
+// OfrepEventStream holds the configuration used to advertise the flag-change SSE
+// endpoint in the OFREP bulk evaluation response (OpenFeature ADR-0008).
+type OfrepEventStream struct {
+	// Enabled turns on the eventStreams field in the OFREP bulk evaluation response.
+	Enabled bool `mapstructure:"enabled" koanf:"enabled"`
+
+	// Endpoint is the public base URL clients should use to reach the relay proxy SSE
+	// endpoint, for example https://gofeatureflag.example.com. The path of the SSE
+	// endpoint is appended automatically. It is required when Enabled is true.
+	Endpoint string `mapstructure:"endpoint" koanf:"endpoint"`
```
thomaspoignant, line 12: It may be better to name this `BaseURL` it is more explicit that we will complete the path with the SSE endpoint.

thomaspoignant (commented):
Thanks a lot @ChrisJr404 for this PR this is really great and needed to support latest OFREP features.

I have put some comments, I am happy to look at them with you if needed.

ChrisJr404:
Thanks for the review. Answers to your points:

- **Enable by default:** I would keep it opt-in for this PR. Advertising the SSE endpoint changes what OFREP providers see in the bulk response, so I would rather not turn it on for everyone until the providers have been exercised against it. Happy to flip the default in a follow-up once that testing is in place, which also gives a clean place to add the provider matrix you mentioned.
- **BaseURL:** agreed, I will rename the field to `BaseURL` since we complete it with the SSE path.
- **Config validation:** I will add validation in config_validator.go so that `enabled: true` with an empty endpoint fails explicitly, and that the endpoint is an absolute URL.
- **apiKey:** good point that the endpoint can need an API key. I will thread that through so it can be included when we answer the provider, per the ADR's opaque-url representation.

I will push the rename plus validation first, then follow up on the apiKey handling.

thomaspoignant:
@ChrisJr404 Do you want me to follow up on this PR or are you still planing to do it?

